PR #10, opened 2020-01-11
Adding Subparsers with Union[...]

lebrice:
Added arparse Subparsers, by specifying an attribute as a Union of dataclass types.

### DIFF
--- a/test/test_subparsers.py
+++ b/test/test_subparsers.py
@@ -119,6 +119,7 @@ def test_command_tree():
     assert prog.command.subcommand.value == "jack"
     assert prog.execute() == "jack"
 
+
 if __name__ == "__main__":
     import sys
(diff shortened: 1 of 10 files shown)


test/test_subparsers.py
```diff
@@ -0,0 +1,62 @@
...
+    metric: str = "accuracy"
+
+@dataclass
+class GlobalOptions(TestSetup):
+    """ Global Options """
+    # mode, either Train or Valid.
+    mode: Union[TrainOptions, ValidOptions] = subparsers({
+        "train": TrainOptions,
+        "valid": ValidOptions,
+    }, default="train")
```
lebrice: Should people be forced to use either subcommand? or is something like "default" okay?

Issue #21, opened 2020-05-15
Fix the Logging config.


lebrice:
**Describe the bug**
Not sure how, but there might be a smarter way to configure the logging so as not to pollute the output of whatever code is using simple_parsing, following [https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)

wolf1986:
Hi! Thanks for the nice library!

I've noticed a problem with logging - it wasn't done the standard way, and was hard to integrate into projects with other loggers.

Added more info & explanations in this PR: #79

PR #23, opened 2020-07-07
Remove call to logging.basicConfig in the module

ivanprado:
Otherwise, code using this library will be forced to use this handler and
it potentially will generate duplicated log lines in the output, as both
the user and these library handlers are installed. This is actually my case. 

basicConfig method should be only called by the library user (e.g. in the main function), and never inside the
module of the library itself. 

This can be seen as a first step for issue https://github.com/lebrice/SimpleParsing/issues/21

### DIFF
--- a/simple_parsing/logging_utils.py
+++ b/simple_parsing/logging_utils.py
@@ -1,12 +1,6 @@
 import logging
 from pathlib import Path
 
-logging.basicConfig(
-    format='%(levelname)-8s [%(name)s:%(lineno)d] %(message)s',
-    datefmt='%Y-%m-%d:%H:%M:%S',
-    level=logging.INFO,
-)
-
 def get_logger(name: str) -> logging.Logger:
         """ Gets a logger for the given file. Sets a nice default format. 
         TODO: figure out if we should add handlers, etc. 


lebrice:
Oops, thanks for pointing this out!
Feel free to merge this.

On a sidenote, would you be interested in sharing a bit about your use-case for SimpleParsing? I'm always looking for some feedback/suggestions on what could be added to it or how it could be improved.

Thanks again!

ivanprado:
> On a sidenote, would you be interested in sharing a bit about your use-case for SimpleParsing? I'm always looking for some feedback/suggestions on what could be added to it or how it could be improved.

@lebrice I was considering to use SimpleParsing for my project, but AFAIR it doesn't support nested lists of dataclasses which was a requirement for my project, so finally I end up implementing something internally. 

But definitely I'm using the cool method `get_attribute_docstring` :smiley:  Very useful to autogenerate the documentation :clap: 

As suggestion for improvements, I think there is room to have better documentation. I found a little bit difficult to understand what was supported and what was not. 

Anyway, it is a cool project. I liked very much your approach. Keep improving it :smiley:

ivanprado:
BTW, I think I don't have the rights to merge it. Could you do it?

lebrice:
@ivanprado  Interesting!
You're right, parsing nested lists of dataclasses **from the command-line** doesn't work, but it does from json/yaml/etc files, whenever the dataclass extends the `simple_parsing.Serializable` class. I don't know if that helps in your particular case, but just thought I'd throw it out there.
 
I'm curious, what do your nested dataclasses represent? How would you expect it to work, roughly?

For example, assuming we have this code:
```python
@dataclass
class ResblockOptions:
    """ Configuration/hparams of a resnet block. """ 
    # Number of filters in the resblock.
    num_filters: int = 64
    # Another parameter for each resblock.
    foo: float = 1.23 

@dataclass
class ResnetOptions:
    """ Configuration options for a Resnet """
    # Options for each resblock. By default uses three resblocks, 64 filters each.
    blocks: List[ResblockOptions] = list_field(Resblock(), Resblock(), Resblock())

parser = ArgumentParser()
parser.add_arguments(ResnetOptions, dest="options")
options: ResnetOptions = parser.parse_args().options
my_resnet = Resnet(options)
...
```
And then from the command-line, how would you expect it to be called?

- `python example.py --num_filters 64 128 128 --foo 1 2 3`?
- `python example.py --num_filters 64 --foo 1 --num_filters 128 --foo 2 --num_filters128 --foo 3`?
- Something else?

Let me know what you think, and thanks again.

ivanprado:
@lebrice In my case, I also need the command-line support. I have some concept of a task and the process accept a list of tasks. 
```py
@dataclass
class Task:
    attr1: int
    attr2: Optional[float] = 2.2
    attr3: Optional[str] = "a string"

@dataclass
class Job:
  tasks: List[Task] = list_field()
```

And the way I submit the tasks is for example:
* `program --tasks.1.attr1 10 --task.1.attr2 0.5 --tasks.2.attr1 3 --tasks.2.attr3 hey`

I'm using `1` and `2` above to group the attributes from different tasks. You can put whatever, as it is used only for grouping in my case. 

That's how I solved, just in case it gives you ideas.

PR #78, opened 2021-07-27
introduce list of choices

mixilchenko:
- allow `is_choice` to work with `is_list`
- fix choice `default` default value from `None` to `MISSING`
  `None` default breaks `default_factory` handling
- remove redundant `test/test_choice.py::test_choice_with_dict_weird`
- add tests with lists of choices

### DIFF
--- a/simple_parsing/wrappers/field_wrapper.py
+++ b/simple_parsing/wrappers/field_wrapper.py
@@ -63,11 +63,6 @@ class FieldWrapper(Wrapper[dataclasses.Field]):
         # stores the resulting values for each of the destination attributes.
         self._results: Dict[str, Any] = {}
 
-        if self.is_choice:
-            choice_dict = self.field.metadata.get("choice_dict")
-            if choice_dict:
-                self._type = str
-
     @property
     def arg_options(self) -> Dict[str, Any]:
         """Dictionary of values to be passed to the `add_argument` method.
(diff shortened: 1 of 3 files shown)


test/test_choice.py
```diff
@@ -48,33 +58,34 @@ class BB(Base):
...
-    c = C.setup("--option bob")
-    assert c.option.value == "bobobo"
+    c = C.setup("")
+    assert c.option == AA("aaa")
+    assert c.options == [AA("aaa")]
 
-def test_choice_with_dict_weird():
-    @dataclass
-    class D(TestSetup):
-        option: List[Base] = choice({
```
mixilchenko, line 65: @lebrice I think this test was incorrect.
Type of `option` must be `Union[List[Base], float, Base]` because we specify here type of final attribute.

I've merged this test with `test_choice_with_dict` above.
If you disagree, I can fix it as you say

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -687,10 +684,15 @@ def required(self, value: bool):
...
         """Returns the wrapped field's type annotation."""
-        if self._type is not None:
-            return self._type
-        else:
+        if self._type is None:
             self._type = self.field.type
+            
+            if self.is_choice and self.choice_dict:
+                if utils.is_list(self.field.type):
+                    self._type = List[str]
```
lebrice, line 692: Interesting, I didn't think of doing it that way.
I think it would ideally be better not to do it through changing `self._type`, because this indirectly depends on the other part of the code that constructs the `arg_options` dict that gets passed to `add_argument`.

For now you can leave this as-is, and I'll think about it some more.

test/test_choice.py
lebrice: Sure, this is fine. Thanks!

lebrice (approved):
Hey @mixilchenko , thanks a lot for making a contribution! :smile: 

This looks good! Feel free to merge this, and I'll push a new tag soon.

Thanks again!

mixilchenko:
@lebrice I don't have write access to this repo so I can't merge this request.
For me it's ok for owner to merge

PR #79, opened 2021-07-29
Fix: Cleaner logging

wolf1986:
The custom method `logging_utils.get_logger()` aimed to organize the logs per file, by using the full file path `__file__`, however it is wrong and is causing a lot of clutter - 

1. In Python, logs are organized by hierarchy - each logger may have a parent logger. when using the default `getLogger('parent_name.child_name')` notation, it automatically creates multiple loggers and assigns parents according to the dots (`.`) in the name. Note that `getLogger` is actually a singleton manager, and can be called multiple times. A logger will be created only if it didn't exist yet. [More info here](https://docs.python.org/3/library/logging.html#logger-objects)

1. The usual case is that every module (defined by a `.py` file) creates a logger - by calling `getLogger(__name__)`. While `__name__` is the name of the module (including packages separated by dots) - thus the hierarchy remains.

1. After the proposed change, every file (module) in `simple-parsing` package, will create its own logger with the correct hierarchy, resulting in a parent `simple-parsing` logger that can be retrieved by anyone, and hushed if desired using `getLogger('simple-parsing').setLevel(logging.WARNING)` - Note that this will set the parent logger, and all the child loggers that report to it - will be also affected this way. **Before this change, there was no way to silence the multiple logs of this package.**

Some suggestions about logging:
1. Every module creates its own logger using `logger = getLogger(__name__)` .
2. Log using the appropriate log-level `logger.debug(...)` ... `logger.warning(...)` - according to the situation.
3. Only the main script (the user of the package, or some high-level entry scripts of the library) is responsible for creating log-handlers and log-formatters. Every main script is responsible to decide what logs to show and what not (for example, silence libraries a, b, c at level WARNING and libraries d, e at level INFO)
4. Log formatting - describes how to output the message of the log - show date/time, show process-id, etc... Log formats support filenames and line-numbers [More info here](https://stackoverflow.com/a/44401529/491570)

In this pull request:
- Modified: Use `logging.getLogger(__name__)` instead of custom `logging_utils.get_logger()`
- Removed: `logging_utils.py` - It is not needed, use a formatter that can show the filename if desired


lebrice (approved):
Awesome! Thanks a lot @wolf1986 !

The reason why I used `__file__` instead of `__name__` is as you mentioned, I wanted to be able to "ctrl+click" on the log to get directly to the line that caused it. As you mentioned, that could better be achieved through a logging formatter, which I didn't know about.

One question then: Is it possible to add such a formatter only to the Logger of simple-parsing? Or would that have an effect globally?

lebrice:
I'll merge this PR right now if that's alright with you! :)

wolf1986:
Thanks that was quick. I'll write a more detailed answer to what you asked later.
Regarding merge - I ran the tests - most of them pass, I think that the failed ones probably failed before my changes.

If this seems ok to you then lets go ahead and merge.
I now see that "Merging is blocked, The base branch requires all commits to be signed". I'll check if I have to change something on my end.

lebrice:
Are some tests failing on your end? That would be odd, since the CI check with travis (which runs pytest in python 3.6, 3.7 and 3.8) is passing!

lebrice:
About the signed commits, you can just `git commit --amend -s` to add a signature to your last commit I think. Otherwise the link that GitHub shows might also be relevant.

wolf1986:
> About the signed commits, you can just `git commit --amend -s` to add a signature to your last commit I think. Otherwise the link that GitHub shows might also be relevant.

Thanks, will try now


> Are some tests failing on your end? That would be odd, since the CI check with travis (which runs pytest in python 3.6, 3.7 and 3.8) is passing!

Oh, I ran them on Windows, might have something to do with that. Travis CI should have the final say on the matter.

lebrice:
Idk why GitHub is still acting up. There's probably some local configuration you need to do. But in the meantime, I'll just overrule this and merge your PR.

wolf1986:
> One question then: Is it possible to add such a formatter only to the Logger of simple-parsing? Or would that have an effect globally?

Didn't try to run the following snippet, but it should be close enough.

```python
    # Handler - handles log-records, responsible to do something with them (e.g. print to stdout)
    # Formatter - applies formatting to log-records using available metadata such as level, time, file/path
    handler = logging.StreamHandler(sys.stdout)
    handler.setFormatter(
        logging.Formatter('{asctime} {pathname}: {levelname} {message}', style='{')
    )

    # logging.getLogger() - Will find the pointer for the logger by the name provided. If this is the first call
    # with the specific name - it will be created.
    # Note: It doesn't matter at all who created it first. The main script will get the pointer and may decide to
    # attach a handler to it. The library - will get the pointer to the same logger, and provide log-records at
    # runtime using e.g. `logger.info()`
    specific_logger = logging.getLogger('simple-parsing')

    # Attach a handler to a specific logger (usually attached to the root logger - the one without a name ''
    # using `basicConfig()`, which only works with the root logger)
    # Note: Multiple handlers can leave peacefully together - root logger can write to both stdout and files,
    # while some specific logger will write to a separate file or a even a log-server
    specific_logger.addHandler(handler)

    logger.info('Hello')
```

PR #82, opened 2021-09-07
Add encoding rule for Namespace

tristandeleu:
- Add an encoding rule for `Namespace` type, in order to directly call `encode` over the arguments parsed by `argparse`/`simple-parsing`.

### DIFF
--- a/simple_parsing/helpers/serialization/encoding.py
+++ b/simple_parsing/helpers/serialization/encoding.py
@@ -17,6 +17,7 @@ from os import PathLike
 from typing import (Any, Dict, Hashable, List,
                     Set, Tuple, Union)
 from collections.abc import Mapping
+from argparse import Namespace
 
 
 logger = getLogger(__name__)
@@ -132,3 +133,7 @@ def encode_dict(obj: Mapping) -> Dict[Any, Any]:
 @encode.register(PathLike)
 def encode_path(obj: PathLike) -> str:
     return obj.__fspath__()
+
+@encode.register(Namespace)
+def encode_namespace(obj: Namespace) -> Any:
+    return encode(vars(obj))


lebrice (approved):
Hey Tristan, thanks for this!
Can you give me an example use-case for this, just out of curiosity?

tristandeleu:
Personally I'd be using it for logging the configuration in Weights & Biases. Usually when I'm using `argparse`, `wandb` directly converts `Namespace` into a dictionary (they most likely simply do `vars(args)`) they can send to the server.

When I'm using `simple-parsing` with wandb, the dataclasses are serialized as a raw string; for example
```json
"metalearner": {
    "desc": null,
    "value": "MAMLArguments(alpha=0.01, num_steps=5)"
}
```

Even though this retains all the information about my configuration (I can simply do `eval(config['metalearner'])` to get the dataclass back), it makes searching on the wandb side impossible (e.g. I can't search by `metalearner.num_steps`).

I was looking at a way to serialize the whole `args` in such a way that it would be wandb-friendly, and I came across the amazing `encode` function, which does exactly that (converts a datastructure, including dataclasses, into a dictionary). This rule is just for convenience to encode the whole `Namespace` at once, without having to `encode(vars(args))`. Concretely, here is what logging the config looks like:

```python
wandb.config.update(encode(args))
```

(I originally intended to have the rule defined in my project, but I figured this might be of general interest)

lebrice:
Interesting! 

Are you using `Serializable` (from simple_parsing.helpers) as a base class for your Config? (I suspect you are, but if not, take a look at [this example](https://github.com/lebrice/SimpleParsing/tree/master/examples/serialization).)

Bonus: Off-topic, but I'd encourage you to check out [the `HyperParameters` class](https://github.com/lebrice/SimpleParsing/blob/master/simple_parsing/helpers/hparams/hyperparameters_test.py), it's the kind of thing I hope you might like.
The idea is to let people write the priors for their hyper-parameters directly in code.  It has nice a integration with `orion` and is easy to log with wandb (as you did above). Providing a tighter integration with wandb is something I think could be interesting.

Let me know what you think :)

PR #86, opened 2021-09-29
Support Literals and Finals in Dataclasses

idoby:
Addresses issue #2


simple_parsing/wrappers/dataclass_wrapper.py
```diff
@@ -77,7 +78,13 @@ def __init__(
                 child_wrapper.required = False
                 child_wrapper.optional = True
                 self._children.append(child_wrapper)
-
+            elif utils.is_final(field.type):
+                # Final fields have no business being set by the command line
+                continue
```
mixilchenko, line 83: I don't think this is dataclass way.
Currently dataclasses do not use `Final` at all.
`Final` means variable should not be changed after initialization.
But it must be initialized so it's not the same as skippable `field(init=False)` or `ClassVar`

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -95,11 +94,11 @@ def arg_options(self) -> Dict[str, Any]:
...
-        self,
-        parser: argparse.ArgumentParser,
-        namespace: argparse.Namespace,
-        values: Any,
-        option_string: Optional[str] = None,
+            self,
+            parser: argparse.ArgumentParser,
+            namespace: argparse.Namespace,
+            values: Any,
+            option_string: Optional[str] = None,
```
mixilchenko, line 101: Your formatter does it wrong. Check for other cases too please

mixilchenko (changes requested):
Lists (and other containers) of literals should be supported and tested too.

simple_parsing/wrappers/dataclass_wrapper.py
idoby: Final fields are being used in classes and dataclasses to express a constant field/member like in typesafe languages such as Java, so I think it makes sense for SP not to do the wrong thing. A Final field must be initialized with a value at the point of declaration, of course, however, the point of the PR was to support a literal as a shorter and more standardized idiom in addition to choice().

If you think SP shouldn't support Finals at all, that's fine by me, but it would do the wrong thing in case a client *does* choose to use Final.

simple_parsing/wrappers/field_wrapper.py
idoby: Please commit a config file for black. I just used pycharm's pep8 based formatting.

lebrice: Tha's a good idea! Don't worry about formatting for now though, I'll take care of that.

simple_parsing/wrappers/dataclass_wrapper.py
lebrice: Let me see if I understand the options here (and please correct me if I'm wrong):

Option 1: (@idoby) Don't create command-line options for Final fields. (Since they are normally used as constant attributes on dataclasses)

Option 2: (@mixilchenko) Create command-line options for Final fields, since the Final annotation means "constant after initialization".

Did I understand that correctly?

simple_parsing/utils.py
```diff
@@ -479,6 +475,24 @@ def is_optional(t: Type) -> bool:
     return is_union(t) and type(None) in get_type_arguments(t)
 
 
+# Note: the typing module is still under development
+# and the way to detect Final and Literal may change in the future
+
+
+def is_final(t: Type) -> bool:
+    if sys.version_info < (3, 8, 0):
+        return False
```
lebrice, line 484: This isn't quite correct! People can use Final and Literal by importing them from the `typing_extensions` package, when their python versions are lower than 3.8.

I think adding something like this at the top of the file might be useful (btw, I wasn't aware of `typing_extensions` when I first wrote this module):
```python
try:
    from typing import Literal, Final, TypedDict # and every function you might need like get_origin etc
except ImportError:
    from typing_extensions import Iteral, Final, TypedDict, # and every function you might need
```

I'm not 100% sure, but I think the `typing_extensions` package might already have all the functions that `typing_inspect` provides.

simple_parsing/utils.py
```diff
@@ -479,6 +475,24 @@ def is_optional(t: Type) -> bool:
...
+
+    return isinstance(t, _GenericAlias) and t.__origin__ == Final
+
+
+def is_literal(t: Type) -> bool:
+    if sys.version_info < (3, 8, 0):
+        return False
```
lebrice, line 491: Same here

simple_parsing/wrappers/dataclass_wrapper.py
idoby: On second thought, I'd prefer option 2 because option 1 views the dataclasses as specific to the command line, which I would like to avoid (I see them as domain-driven objects that SP lets me conveniently pull from the command line), but this is confusing so we'll need to document this very very well.

simple_parsing/wrappers/field_wrapper.py
idoby: I formatted with black's defaults, not too many things have changed so I guess it's mostly the same config.

simple_parsing/wrappers/dataclass_wrapper.py
```diff
@@ -77,7 +80,23 @@ def __init__(
...
+            elif utils.is_literal(field.type):
+                choices = field.type.__args__
```
lebrice, line 87: Does this check for an `__args__` attribute work in all python versions (3.6+?)
I think it might be preferable to use a function for this if not, something like `get_type_arguments` (that I think I've already written somewhere in utils), or maybe a better function from `typing_extensions`.

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -164,61 +169,27 @@ def get_arg_options(self) -> Dict[str, Any]:
             # automatically adds the (default: '123')
             _arg_options["help"] = " "
 
-        if utils.is_choice(self.field):
-            _arg_options["type"] = str
-            _arg_options["choices"] = list(self.choices)
+        if self.is_choice:
+            if isinstance(self.choices, dict):
+                _arg_options["type"] = str
+                _arg_options["choices"] = list(self.choices.values())
```
lebrice, line 175: Assuming from memory that `self.choices` was a dict before as well, here you're changing `arg_options["choices"]` from from being the keys of `choices` to instead be the values, right? Doesn't this cause any issues in tests?

idoby:
I added support for lists, tuples and optionals of literals, but supporting anything with a more complicated structure is 1) dubious and 2) will require a refactor to some of functions in the field wrapper to do easily and without bugs.

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -164,61 +169,27 @@ def get_arg_options(self) -> Dict[str, Any]:
...
+            else:
+                assert len(set(type(choice) for choice in self.choices)) == 1
```
lebrice, line 177: Here you're asserting that all the options are of the same type, right? I agree, fields like `Literal["cat", 1.0, True]` probably shouldn't be supported. However, this should raise some type of error with a message explaining this.

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -1042,3 +997,47 @@ def get_argparse_options_for_annotation(t: Type) -> Dict[str, Any]:
     #     else:
     #         _arg_options["nargs"] = "*"
     return _arg_options
+
+
+def _optional_args(_arg_options, type_arguments):
```
lebrice, line 1002: I agree that the code block for optional args was a bit ugly, but I don't think it's super relevant to this PR to move it here. Plus, I'm not really able to tell if anything was changed in that block. Would you mind reverting this?

test/test_final.py
```diff
@@ -0,0 +1,25 @@
+from dataclasses import dataclass
+
+from .testutils import *
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 8), reason="Final is only supported in Python >= 3.8"
```
lebrice, line 7: See previous comment: This test should work in all versions, but use the Final from `typing_extensions` if python < 3.8

test/test_literal.py
```diff
@@ -0,0 +1,165 @@
...
+
+from .testutils import *
+
+
+# Not using simple_attribute because floats are disallowed in Literals
+# and booleans don't make sense in this context
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 8), reason="Literal is only supported in Python >= 3.8"
```
lebrice, line 11: Same here, use `Literal` from `typing_extensions` if Python < 3.8.

test/test_literal.py
```diff
@@ -0,0 +1,165 @@
...
+)
+def test_literal_argument_str():
```
lebrice, line 13: Here you should instead take advantage of `SomeClass` being an inner class in order to do something really powerful:


```python3
@pytest.mark.parametrize(
    "allowed_values",
    [
        [123, 456],
        [123],
        ["a", "b"],
        [True, False],
        # etc etc
)
def test_literal_argument(allowed_values: List[Any]):
    @dataclass
    class SomeClass(TestSetup):
        a: Literal[*allowed_values]
    parser = ArgumentParser()
    parser.add_arguments(SomeClass, dest="some_class")
    for allowed_value in allowed_values:
        args = parser.parse_args(shlex.split(f"--a {allowed_value}"))
        assert args.some_class.a == allowed_value
```

Or something like this. Basically what I'm saying is, type annotations are funky, and you can parametrize them (that's kind-of what I'm doing in the above test with `simple_type`)

lebrice (changes requested):
Hey @ivanprado, Thanks for this contribution, awesome stuff!

There are a few things to address, but this should't be hard. Let me know if you have any questions/comments.
Thanks again!

simple_parsing/utils.py
idoby: Yeah, I didn't know if we wanted to bring in a dependency on typing extensions. I can add it if you want.

simple_parsing/wrappers/dataclass_wrapper.py
idoby: Thanks, I just found that function while adding support for lists.
Well, the tests did pass with all versions, so it seems to be good, but switching to the function (which does the same basically) is a good idea anyway. :ok_hand:

simple_parsing/wrappers/field_wrapper.py
idoby: self.choices is now passed in from DataclassWrapper. Basically I think this code would be a lot cleaner and more flexible if it acted solely as a mechanism, leaving policy decisions to DCW. So at this point, FieldWrapper doesn't care if the choice froms from Literal, choice() or anywhere else. Maybe we can leverage the same code for enums too later on, but this is outside the scope of this PR.

simple_parsing/wrappers/field_wrapper.py
idoby: Technically the PEP allows mixing different types in the Literal declaration, but I couldn't find a way to express this to argparse. Maybe you know how to do this.

simple_parsing/wrappers/field_wrapper.py
idoby: I'm using it in two places now, would you prefer it if I duplicated the code?

It is hard to see the changes but basically there are none (except for passing in type_arguments as a parameter)

test/test_literal.py
idoby: I would but parameterizing Literal with anything other than a literal is not allowed by the PEP. We can do this, but then the test itself would not be PEP-compliant. I also seem to recall this didn't work for some reason.

What would you prefer?

simple_parsing/wrappers/dataclass_wrapper.py
lebrice: Also, here's one good thing about option 2: creating the arguments for a field like  `a: Final[SomeType]` will be exactly the same as creating the arguments for a `a: SomeType` field!

simple_parsing/utils.py
lebrice: It's probably already installed because of typing_inspect I think

simple_parsing/wrappers/field_wrapper.py
lebrice: Sure, I agree, but that's not what I'm asking exactly: I'm asking, say the choices was a dict with a non-trivial type for the values, say something like:
```python
optimizer: Type[torch.optim.Optimizer] = choice({"adam": torch.optim.Adam, "sgd": torch.optim.SGD})
```
I'm asking if this change here will make the options to choose from, from the command-line, to be `"{<class Adam> <class SGD>}"` instead of just `{"adam", "sgd"}`

simple_parsing/wrappers/field_wrapper.py
lebrice: Nono it's fine for Simple-Parsing to choose not to support these kinds of fields, even though it's possible for them to exist.
This just needs to raise an error with a good message so that the users know that's not supported.

simple_parsing/wrappers/field_wrapper.py
lebrice: Oh alright then. Didn't spot that.

test/test_literal.py
lebrice: I don't mind the test not being pep-compliant. It's not "real" code anyways, it's more like "meta-code" anyways, code generating what would be pep-compliant code! :)
I think this should work, I've done similar kind of hacks with the Union type elsewhere. Let me know if it doesn't though.

simple_parsing/wrappers/dataclass_wrapper.py
idoby: Well, you'd still need to extract the actual type out of the Final but yeah it's a very simple change

simple_parsing/wrappers/field_wrapper.py
idoby: Good catch, my bad, this does change the semantics. However, the dict field ends up getting overwritten by custom_args. Either way I decided to put all the code in FieldWrapper instead of splitting it. I'd still do it the way I suggested but that would require a refactor/rewrite and I don't want to end this PR with a mixed paradigm which is worse.

lebrice:
> Great work to support python 3.9!
> 
> Lists (and other containers) of literals should be supported and tested too.

Huuh just for future reference @mixilchenko , I don't quite agree with the second point.
Adding Literal and Final is already a nice, simple contribution. Adding lists and tuples of finals seems more like a bonus, somewhat niche use case.
Sorry for noticing this a bit late @ivanprado, feel free to add or not add the containers of Finals as you wish. If not, then we can probably do that in another PR. If things are already working on your end, then fine, this PR can also add them.

simple_parsing/utils.py
lebrice: There should also be a test for `is_final` and another for  `is_literal`. Something like this:
```python
@pytest.mark.parametrize("some_type, expected",
    [
        (Final, True),
        (Dict, False),
        (Final[str], True),
        (Tuple[str], False),
        ... # etc etc
    ]
)
def test_is_final(some_type: Type, expected: bool):
     assert is_final(some_type) is expected
```

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -1042,3 +1022,62 @@ def get_argparse_options_for_annotation(t: Type) -> Dict[str, Any]:
...
+
+def _literal_args(_arg_options, t):
+    type_arguments = utils.get_type_arguments(t)
+
+    if len(set(type(arg) for arg in type_arguments)) != 1:
+        raise ValueError("Only Literals of the same type are supported")
+
+    literal_type = type(type_arguments[0])
+
+    if literal_type not in [str, int, bool, bytes] and utils.is_enum(literal_type):
```
lebrice, line 1079: Wait, you're saying that enums are supported in the string below, but your test condition above uses `and utils.is_enum`, did you mean to write this instead?

```suggestion
    if literal_type not in [str, int, bool, bytes] and not utils.is_enum(literal_type):
```

And if so, then ideally some test should have been able to catch that!

test/test_literal.py
```diff
@@ -0,0 +1,174 @@
...
+        ),
+        xfail_param(
+            format_list_using_single_quotes,
+            reason="TODO: decide which syntax we want to allow for single lists.",
+        ),
+    ],
+)
+def test_literal_argument_list(list_formatting_function):
+    @dataclass
+    class SomeClassInt:
```
lebrice, line 121: Please use the `TestSetup` class as a base class for your inner classes. This will make all the checks everywhere a lot easier to read:
Before:
```python
parser = ArgumentParser()
parser.add_arguments(SomeClassInt, dest="some_class")
args = parser.parse_args(shlex.split(f"--a 123"))
assert args == argparse.Namespace(some_class=SomeClassInt(a=[123]))
```
After:
```python
assert SomeClassInt.setup("--a 123") == SomeClassInt(a=[123])
```

lebrice (changes requested):

lebrice:
Thanks again @ivanprado , I added some more suggestions.

lebrice:
Hey @ivanprado, just in case this looks a bit overwhelming:
I'd be happy to actually implement/fix most of my comments here myself, if you don't have time, but I'd need write access to your fork, so that I can push new commits to your branch.

Let me know if that's something you'd like :)

lebrice:
Woops you're right, sorry about that!

idoby:
I forgot about this PR. Should I close it?

lebrice:
I mean, it would be nice to have this in, actually! But probably best to close this and start fresh with my comments in mind.

Up to you @idoby .

PR #87, opened 2021-09-29
Add Python 3.9.7 to CI testing + fix tests to work in all environments

idoby:


mixilchenko (changes requested):
@lebrice reformatted code with black in #85 
I think it would be great if you do the same with your code

idoby:
Reformatted

examples/ML/other_ml_example.py
```diff
@@ -53,9 +51,10 @@ class EvalConfig:
 train_config: TrainingConfig = args.train_config
 eval_config: EvalConfig = args.eval_config
 
-print(hparams, train_config, eval_config, sep="\n")
-expected = """
-MyModelHyperParameters(batch_size=32, learning_rate=0.05, momentum=0.01)
-TrainingConfig(data_dir='data', log_dir='logs', checkpoint_dir='checkpoints')
-EvalConfig(eval_dir='eval_data')
-"""
+assert hparams == MyModelHyperParameters(
```
lebrice, line 54: This isn't quite right, but I guess it wasn't clear why I was doing it that way to begin with, so totally understandable.

The point here is that the examples, when run with no arguments, should give `expected`. This is what the `test_examples` script checks for.
But we still want the examples to be run with different arguments! Here this will raise an AssertionError, which we don't want!

Please revert these types of changes.

examples/dataclasses/hyperparameters_example.py
```diff
@@ -28,21 +28,15 @@ def load(cls, path: str):
             return cls(**config_dict)
 
 
+parser = ArgumentParser()
 parser.add_arguments(HyperParameters, dest="hparams")
 
 args = parser.parse_args()
 
-hparams: HyperParameters = args.hparams
```
lebrice, line 35: Do you avoid doing this type of unpacking because of your type checker? They are fine with MyPy, which I use. I'd rather keep those as they are.

test/test_examples.py
```diff
@@ -2,118 +2,28 @@
...
-        out_lines = out.splitlines(keepends=False)
-        out_lines = [line.strip() for line in out_lines if line and not line.isspace()]
-        expected_lines = expected.splitlines(keepends=False)
-        expected_lines = [
-            line.strip() for line in expected_lines if line and not line.isspace()
-        ]
-        assert out_lines == expected_lines, file_path
 
-    return should_equal
+def pytest_generate_tests(metafunc):
```
lebrice, line 14: As mentioned above, while this is clever from a testing point of view, I don't want assertions in the examples. Please revert all of this.

test/test_issue_46.py
```diff
@@ -18,10 +22,11 @@ def test_issue_46(assert_equals_stdout):
 
     parser.print_help()
 
+    running_name = os.path.basename(sys.argv[0])
```
lebrice, line 25: Just out of curiosity, in what context is this not `pytest`?

lebrice (changes requested):
Hey @ivanprado , thanks for this.

I'm sorry to say that most of this won't make it in: I really don't want assertions in the examples, otherwise users can't play around with them by passing other values.

Which tests fail when you just add the `3.9` option?

examples/ML/other_ml_example.py
idoby: I thought you would say this. The reason I changed these tests is because they fail with 3.9, due to fields being printed in a different order. Fixing this would insert unrelated boilerplate into the example code.

A compromise could be not to try to check the output of the examples in automated tests. I think this is fine because there are plenty of actual tests for this library.

What do you think?

examples/dataclasses/hyperparameters_example.py
idoby: TBH I changed this because I thought they added a redundant line to the example, diluting the point that the user should pull the objects from the args object. It's a matter of taste, I guess. I'll revert these changes.

examples/ML/other_ml_example.py
lebrice: Ohh the ordering of the fields is different in 3.9? that's interesting! Do you know what caused that change, out of curiosity?
Simplest solution to me seems like just xfail-ing the test in `test_examples.py` for python 3.9.
Another idea could be to just check that the word counts match instead of checking for equality when using python 3.9, and keep checking for equality with other versions!
Something like which would make the test order-agnostic, just for python 3.9:
```python3
from collections import Counter
assert Counter(actual_string.split()) == Counter(expected_string.split())
```

test/test_issue_46.py
lebrice: Ok cool!

examples/ML/other_ml_example.py
idoby: I vaguely recall something about the internal implementation of dictionaries changing, and who knows what part of the code we're using here, is built on top. But I didn't look into it in detail really.

Counting words just means relying on strings again, and I'm not really sure this provides any value. I'd rather just skip the tests for 3.9. The examples are mostly simple and are covered by proper tests anyway.

test/test_examples.py
```diff
@@ -78,19 +48,11 @@ def test_running_example_outputs_expected(
...
-        # get the 'expected'
-        # assert "expected" in resulting_globals
-        if "expected" not in resulting_globals:
-            pytest.xfail(reason="Example doesn't have an 'expected' global variable.")
-        expected = resulting_globals["expected"]
-        assert_equals_stdout(expected, file_path)
 
-    except SystemExit as e:
+    except SystemExit:
         pytest.xfail(f"SystemExit in example {file_path}.")
```
lebrice, line 58: Adding something like this here (and reverting the above changes) will be fine for now:
```suggestion
@pytest.mark.xfail(sys.version_info < (3, 9), reason="ordering of fields is different in python 3.9")
```

lebrice (changes requested):

lebrice:
Closing this, I ended up adding python 3.9 myself in #110

PR #94, opened 2021-10-12
Intuitive boolean flags

### LINKED ISSUE #68: Confusing command line options for bool's with `default=True`
Having a boolean option with a default value of `True`:
```

from simple_parsing import ArgumentParser


@dataclass
class Args:
    enable_water: bool = True  # Do you want water?


parser = ArgumentParser()
parser.add_arguments(Args, dest="args")

print(parser.parse_args(["--enable_water"]))
# => Namespace(args=Args(enable_water=False))
```

Creates this very confusing behaviour that passing the option (without a boolean value) disables it, so passing `enable_water` actually causes `enable_water` to be set to False, which is not very intuitive.

Even though this behaviour corresponds to `store_false` in the classic python `ArgumentParser`, it would be nicer to make this more clear. For example by changing the command line option automatically to `--no_enable_water` (by prepending a `no_`) when the default is `False`, and then setting `enable_water` to False when the option `--no_enable_water` is passed.

haydenflinner:
Fixes #68 

( I will rebase this shortly to include only the relevant bool change )

### DIFF
--- a/simple_parsing/wrappers/field_wrapper.py
+++ b/simple_parsing/wrappers/field_wrapper.py
@@ -270,10 +270,8 @@ class FieldWrapper(Wrapper[dataclasses.Field]):
                 _arg_options["type"] = type_fn
 
         elif utils.is_bool(self.type):
-            _arg_options["type"] = utils.str2bool
-            _arg_options["type"].__name__ = "bool"
-            _arg_options["nargs"] = "?"
-
+            _arg_options["action"] = BoolAction
+            _arg_options["metavar"] = '[bool]'
         else:
             # "Plain" / simple argument.
             # For the metavar, use a custom passed value, if present, else do
@@ -384,16 +382,6 @@ class FieldWrapper(Wrapper[dataclasses.Field]):
                 return tuple(raw_parsed_value)
 
         elif self.is_bool:
-            # print(self.name, raw_parsed_value)
-            if self.dest_field:
-                # TODO: This isn't used anywhere, what was the idea again?
-                # other_default = self.dest_field.field.metadata.get("_original_default")
-                pass
-                # print(other_default)
-
-            if raw_parsed_value is None and self.default is not None:
-                logger.debug("value is None, returning opposite of the default")
-                return not self.default
             return raw_parsed_value
 
         elif self.is_list:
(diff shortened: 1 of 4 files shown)


haydenflinner:
**How does using this BoolAction as the action change how the boolean arguments appear in the "--help" text on the command-line?** 
```
usage: adv-ls.py [-h]  [--follow_symlinks [None]]

optional arguments:
  -h, --help            show this help message and exit

ListDirectoryArgs ['args']:
  Specs for list-dir.

  --follow-symlinks [None], --no-follow-symlinks [None]
                        (default: True)
```

**If it does change how they appear, then how come the test for the examples haven't picked up on that? Is it because the FieldWrapper is already correcting the metavar to be bool or something like that?**

It happens that none of your fields in the tests that test help output are booleans :smile:  at least as far as I can tell.

**If we go the route of auto-generating the "--no-" variants, would we need to also create these arguments using a mutually exclusive group? (I'm really not sure about this though, maybe this is already taken care of by argparse when you pass multiple option_strings for a single argument.)**

I'm not sure that it's necessary; it seems that the last flag wins which is fair enough IMO.
```
python ~/code/python/simpcli3/doc/examples/adv-ls/adv-ls.py --no-follow-symlinks --follow-symlinks --no-follow-symlinks True --follow-symlinks --no-follow-symlinks
Received args: ListDirectoryArgs(paths=[], exclude=[], print_format=<PrintFormat.PRETTY: 2>, follow_symlinks=False)
```

JamesDeAntonis:
Is this PR close to mergeable?  I'm dealing with the same problem.  To be clear, the implemented behavior would make it so that booleans have "store_true" regardless of the default value?

lebrice:
> Is this PR close to mergeable?

Hey there @JamesDeAntonis ! Not quite. I'd actually be tempted to close this old PR. I don't think a custom action class for the boolean args is necessary. There's probably a much cleaner and easier way to do this.

>  I'm dealing with the same problem. To be clear, the implemented behavior would make it so that booleans have "store_true" regardless of the default value?

Yes, the idea is that the boolean fields would always have store_true. I was, at some point, successfully convinced that that makes sense, even for arguments like "--no-val" and such.

In the meantime, the best work-around that I've found is to just use a `flag` (from simple_parsing.helpers.fields), which sets the nargs to 1, instead of "?", so that the arguments are more explicit.

I'll take a closer look at this sometime soon. Also, please feel free, if you'd like to take a crack at it, or if you have some ideas, let me know what you think! :)

JamesDeAntonis:
Thanks for the response!

> sets the nargs to 1

I like this idea, I'll use it

PR #113, opened 2022-03-22
Fix parsing defaults for nested dataclasses

iamgroot42:
When parsing a dataclass that has other dataclass fields in turn, and a default dataclass instance is provided, the parser in its current form ignores values for children dataclasses if they are default, even if the given configuration explicitly provides those values.

For instance,

```python

@dataclass
class A:
  p: int
  q: float

@dataclass
class B:
  x: int
  y: Optional[A] = None
```

For this configuration, a dataclass corresponding to the following configuratiton:
```python
{'B': 'x': 3, 'y': {p: 4, q: 0.1}}
```
when parsed by the module, will result in
```python
{'B': x}
```

### DIFF
--- a/simple_parsing/parsing.py
+++ b/simple_parsing/parsing.py
@@ -99,7 +99,8 @@ class ArgumentParser(argparse.ArgumentParser):
         pass
 
     @overload
-    def add_arguments(self,
+    def add_arguments(
+        self,
         dataclass: Dataclass,
         dest: str,
         prefix: str = "",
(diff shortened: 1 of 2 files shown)


lebrice:
Hey there, thanks a lot for the PR!
I've added two tests for this, and also a fix for the `Optional[A]` case, you gave above.
I will create a new PR that will include your original message and commit, with my additional commit(s) on top. Is that ok with you? Or would you rather I make a PR to this branch, and then after you accept it, we merge your PR to master?

iamgroot42:
@lebrice thanks! Either is fine by me :)

Issue #75, opened 2021-07-23
Representing nested args by their full "nesting path" even with no conflicts


Ozziko:
`simple-parsing` is simply awesome (thanks!). I'd like to suggest a feature that is crucial for parsing nested args, in my opinion, and the support of other researchers in my group.  Here is a minimal example to make things clear:
```
from dataclasses import dataclass
from simple_parsing import ArgumentParser, ConflictResolution

@dataclass
class WandbArgs:
    dir: str = ''  # Wandb run cache dir, relative to project_path; leave empty for default
    use: bool # Whether to use wandb for experiment control, logging, saving...

@dataclass
class Data:
    dir: str = ''  # Data dir, relative to project_path; leave empty for default

@dataclass
class ProjectArgs:
    data: Data = Data()
    wandb: WandbArgs = WandbArgs()


parser = ArgumentParser(conflict_resolution=ConflictResolution.EXPLICIT)
parser.add_arguments(ProjectArgs, dest='args')
args = parser.parse_args().args
```
A command line to set all values can be: `--use=True --wandb.dir='wandb' --data.dir='data'`.

Since I passed ConflictResolution.EXPLICIT:
* Nested args with conflicts - args that appear under different dataclasses, `dir` in this example, are allowed and accessed in command line by their full "nesting path", `wandb.dir` and `data.dir`. This is organized and makes sense. 
* HOWEVER, unique args (no conflicts) appear without their "nesting path", in this example `use`. 

**I claim that having `wandb.use` in the command line is much more clear, unambiguous, simple and intuitive than `use` (current behavior), especially for those who have many args and invested time in organizing them with nesting, to make it organized, clear and simple!**.

I found a hack to get this functionality: I force duplication for all args by adding `parser.add_arguments(ProjectArgs, dest='duplicated')` to my example above (and ignoring the duplicated args) - forcing all args to apepar in command line with their full path. However, since I have many args in my current project, I need to also download and modify `simple-parsing` source code - set `max_attempts=200` in `simple_parsing -> conflicts.py` - otherwise it raises an error, assuming so many conflicts are not possible and must be an error.

Therefore I suggest to add a `force_explicit` argument to ArgumentParser, such that `ArgumentParser(force_explicit=True)` makes all nested args appear in command line with their full "nesting path", no hack, no `ConflictResolution.EXPLICIT`, no max number of conflicts - simple, organized and clear!

Thanks!

lebrice:
[...]

PR #117, opened 2022-04-08
force_full_path option to use always full paths to fields as option string

ivanprado:
This PR is an attempt to provide the functionality at https://github.com/lebrice/SimpleParsing/issues/75

I think this functionality is really useful. Indeed, in my particular use case, it is crucial. 

Probably I'm missing something in the code. So I'm very happy to receive feedback about it :-)


simple_parsing/wrappers/field_wrapper.py
```diff
@@ -19,6 +19,19 @@
 logger = getLogger(__name__)
 
 
+class FullPathMode(Enum):
+    """
+    Whether or not to force the full path to be used for the option string.
+    When set to FullPathMode.DISABLED (default), the option string is as
+    flatten as possible. When set to FullPathMode.FULL, the option string
+    is the full path to the field. Sometimes is desired to ignore the
+    first level. Use FullPathMode.FULL_WITHOUT_ROOT in such cases.
```
lebrice, line 28: Would you mind re-organizing this a little bit into more of a bullet-point list for each option, and using the same example argument to illustrate each option? Like "--baz", "--foo.bar.baz", "--bar.baz"?

Alternatively, you could add a "docstring" under each enum value with a description.

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -66,6 +79,9 @@ class FieldWrapper(Wrapper[dataclasses.Field]):
     # Whether to add the `dest` to the list of option strings.
     add_dest_to_option_strings: ClassVar[bool] = True
 
+    # Whether to force all options to use its full path
+    force_full_path: ClassVar[FullPathMode] = FullPathMode.DISABLED
```
lebrice, line 83: This is more of a choice between 3 values than a boolean, so something like `nested_options_mode` or similar would be a preferable name IMO.

lebrice (changes requested):
Hey there @ivanprado , thanks for the PR! 

I've got a few questions/comments:
- Can you give me an example of what kind of use-case there would be for `FULL_WITHOUT_ROOT`?
- I think the `add_dest_to_option_strings` argument is overlapping with this new option, and we should probably fuse them both into a new option instead.
- This deserves a few simple unit-tests

Thanks again for the PR, and let me know what you think!

ivanprado:
Thank you @lebrice . I'm going to work on your feedback. Meanwhile, here I have the answer to your question:

> Can you give me an example of what kind of use-case there would be for FULL_WITHOUT_ROOT

Imagine you want to control all the argument parsing using dataclasses for a server that is serving a machine learning model. The following could be the options dataclasses:

```py
@dataclass
class ModelOptions:
  path: Path
  device: str


@dataclass
class ServerOptions:
  host: str
  port: int
  model: ModelOptions

# There is a function that receives the ServerOptions and starts a server
def start_server(options: ServerOptions):
  ...
```

The default simple-parsing generated arguments would then be `--host str --port int --path Path --device str`. Note how `path` is ambiguous here. Path to what? We could use the new option to create nested args instead using the option `FullPathMode.FULL. The arguments would then be:

```
--options.host str --options.port int  --options.model.path Path --options.model.device str
```

Now the `path` is not ambiguous now. It is clearly the path to the model. But all arguments contains now the redundant word `options`. This is where the option `FullPathMode.FULL_WITHOUT_ROOT` comes to the rescue. It generates the following arguments:

```
--host str --port int --model.path Path --model.device
```

The redundant `options.` prefix has been removed. The first level is plain, and the rest of the levels contains the path of the argument. In summary:

* It is useful to delegate to a single dataclass all the options handling, because then you can pass the parsed object around (in the example we passed `options` to the function `start_server` "as is") 
* But this would generate a redundant prefix for all arguments. The option `FULL_WITHOUT_ROOT` fixes this problem.

ivanprado:
> I think the add_dest_to_option_strings argument is overlapping with this new option, and we should probably fuse them both into a new option instead.

@lebrice I've fused `add_dest_to_option_strings` with this new feature. So now this is control with to arguments:
* `arguments_generation_mode`: controls if arguments must be flat, nested or both
* `nested_mode`: controls how nested arguments should be generated. 

Could you have a look? Thanks you :-)

simple_parsing/parsing.py
```diff
@@ -49,20 +51,33 @@ def __init__(
...
+        - nested_mode : NestedMode, optional
+
+            How to handle argument generation in for nested arguments
+            in the modes ArgumentGenerationMode.NESTED and ArgumentGenerationMode.BOTH.
+            In the NestedMode.DEFAULT mode, the nested arguments are generated
+            reflecting their full 'destination' path from the returning namespace.
+
+            In the NestedMode.WITHOUT_ROOT, the first level is removed. This is useful
+            because sometimes the first level is uninformative. For example,
+            'args'  is redundant and worthless for the arguments
```
lebrice, line 76: It would only be redundant and uninformative if you added arguments for a single all-encompassing dataclass. Maybe just reword this a little bit? Since from reading this, new users could think that `args` will always be the root.

lebrice: Something like "This is useful when `parser.add_arguments` is only called once,  and where the same prefix would be shared by all arguments."

test/test_generation_mode.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+@dataclass
+class ServerOptions(TestSetup):
+  host: str
+  port: int
+  model: ModelOptions
+
+
+def assert_as_expected(options: ServerOptions):
```
lebrice, line 23: Would you mind defining an `expected` variable in the global scope instead, and doing `assert <thing> == expected` instead of calls to this?
```python
expected = ServerOptions(host="myserver", port=80, model=ModelOptions(path="a_path", device="cpu")
```

lebrice (approved):
Thanks @ivanprado , really nice!

I left some minor comments below.

Thanks again!

simple_parsing/parsing.py
ivanprado: @lebrice it makes sense. Addressed here: https://github.com/lebrice/SimpleParsing/pull/117/commits/a7f72239d8a2ec0f7d17caff8225a320cd1de6de

test/test_generation_mode.py
ivanprado: It makes sense. Done here: https://github.com/lebrice/SimpleParsing/pull/117/commits/367b5bc3fbb3c512fda27d2492a0fc491c43dc76

ivanprado:
@lebrice I've addressed your comments. Ready for a re-review.

lebrice (approved):
Thanks again!

ivanprado:
@lebrice thank you to you for accepting my contribution :-)

PR #123, opened 2022-04-09
Removed unnecessary state in ConflictResolver object.

idoby:


lebrice (approved):
Neat, thanks for the PR!
It might be better to just make all the methods static, and then at that point, just removing the class itself becomes easy. The `conflict_resolution` can be passed as an argument to `resolve`, which would be the only public function in that module.

If there are any references to the `ConflictResolver` anywhere, they can be converted to references to the module instead.

But this can be in another PR, no worries.

Thanks!

idoby:
Yeah I thought as much but if you're doing Java-style OOP then having a class with "static" functions is not so bad, and can be considered tidier. Your decision.

lebrice:
I'm fine with this for now. Thanks for the help!

PR #124, opened 2022-04-09
argparse.SUPPRESS-like functionality for advanced uses

idoby:


lebrice:
Interesting!
Can you give me an example use-case for this? And how is this different than just parsing the class and then using `dataclasses.asdict` on the result?

idoby:
The use case is somewhat esoteric, I admit, but this is a feature that is available in argparse (and therefore SP) add_argument, so now it can be used with add_arguments as well.

I'm combining configs from multiple sources, so I need to know exactly what was specified on the command line, since the command line is neither the top priority or lowest priority source I have.
This way I can enable specifying each part of the dataclass from a different source and assemble the DC at the end of the entire process.

It sounds complicated but it makes keeping track of my ML experiments super easy.

lebrice (approved):
Looks fine to me. Let's see if all the tests pass.

Thanks for the PR!

idoby:
(It also affects whether the config MUST be specified on the cmdline, help text etc)

idoby:
I want to test it a little bit more with nested DCs before merging this

idoby:
OK seems to work fine. @lebrice I wrote a small hack when dealing with the boolean test cases there, is there a better way to accomplish this?

test/test_default_args.py
```diff
@@ -81,3 +82,103 @@ class SomeClass:
...
+
+    @dataclass
+    class SomeClass:
+        a: some_type
+        b: int = 5
+
+    parser = ArgumentParser()
+    parser.add_arguments(SomeClass, dest='some_class', default=argparse.SUPPRESS)
+    if some_type == bool:
+        def strbool(s):
```
lebrice, line 124: Use str2bool, wherever it's defined (probably in utils.py)

test/test_default_args.py
```diff
@@ -81,3 +82,103 @@ class SomeClass:
...
+            if s.lower() in ['false', 'no', 'f']:
+                return False
+            else:
+                return True
+
+        parser.add_argument('--c', type=strbool, default=argparse.SUPPRESS)
+    else:
+        parser.add_argument('--c', type=some_type, default=argparse.SUPPRESS)
```
lebrice, line 132: Can you mix suppressed and non-suppressed arguments? What happens in that case?

lebrice (approved):
Seems fine so far, and since it doesn't make any test fail, I don't have any objections to adding this.

Thanks @idoby !

idoby:
@lebrice I'm pretty happy with this now, I'd appreciate your review.

test/test_default_args.py
idoby: I've added another test case to demonstrate this

lebrice (approved):
Looks great! One last thing I think might be worthwhile to test would be to add a suppressed dataclass using `add_arguments`, as well as another dataclass with a non-suppressed default using `add_arguments`.

I don't think it's totally necessary for this to be merged, but if you find the time, I'd also appreciate a test for conflicts between two such argument groups (i.e. doing `parser.add_arguments(SomeClass, "foo", default=argparse.SUPPRESS)`, followed by `parser.add_arguments(SomeClass, "bar")` .

I believe this should bring out some bugs. Then again, you don't _have_ to add this test, but if you're willing to do it, I'd appreciate it.

Thanks again :)

idoby:
> add a suppressed dataclass using `add_arguments`, as well as another dataclass with a non-suppressed default using `add_arguments`.

> a test for conflicts between two such argument groups (i.e. doing `parser.add_arguments(SomeClass, "foo", default=argparse.SUPPRESS)`, followed by `parser.add_arguments(SomeClass, "bar")` .

Rebased onto master and added these tests. No bugs seem to have emerged this time.

idoby:
We can have a zoom call and I'll show you :)

PR #128, opened 2022-04-27
Include requirements.txt

thewchan:
Currently requirements.txt is not included in the sdist, which causes setup.py to fail during building downstream in conda forge

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,3 @@
 include versioneer.py
 include simple_parsing/_version.py
+include requirements.txt


lebrice:
Hey there, thanks for this PR!

>  which causes setup.py to fail during building downstream in conda forge

Interesting! I'd be curious, would you mind elaborating a little bit?
Apologies for the question, since I don't know much about Conda packaging, but are you building simple-parsing through Conda somehow? Or just pip-installing it inside a conda environment?

lebrice (approved):

thewchan:
@lebrice Sure thing! I'm maintaining simple parsing on the conda-forge repo; it basically fetch the sdist from pypi, and run setup.py to build the package, because of that, it needs all the required file in the sdist bundle. E.g., setup.py reads `readme.md` in one of the steps, which works fine since `readme.md` is present in the sdist file bundle on PyPI.

PR #129, opened 2022-04-29
Fix a bug introduced in #117. It affects the DASH option in the nested mode

ivanprado:
Fix DASH option not working in the nested mode


lebrice (approved):
Thanks @ivanprado !

PR #156, opened 2022-08-11
Partials - Dynamic Config Dataclasses for arbitrary callables

lebrice:
This PR adds Partials: Dynamically created dataclasses for arbitrary callables.

It's still a work in progress. I'll update this whenever I find the time.


simple_parsing/helpers/partial.py
```diff
@@ -0,0 +1,308 @@
...
+            isinstance(arg, Hashable) for arg in kwargs.values()
+        ):
+            return cached_fn(*args, **kwargs)
+        return fn(*args, **kwargs)
+
+    return _switch
+
+
+@cache_when_possible()
+def config_dataclass_for(
```
JesseFarebro, line 112: Do you think this is a little verbose? `sp.config` might be a little too opaque, but I'd lean towards `sp.config`.

simple_parsing/helpers/partial.py
```diff
@@ -0,0 +1,308 @@
...
+            current_arg, arg_description = stripped.split(":", maxsplit=1)
+            parsed[current_arg] = arg_description.lstrip()
+        elif line_indent > arg_block_indent:
+            parsed[current_arg] += f" {stripped}"
+    return parsed
+
+
+try:
+    # This only seems to be necessary for the SGD optimizer.
+    from torch.optim.optimizer import _RequiredParameter
```
JesseFarebro, line 234: What's the story here? There's a single exception only for `torch.optim`?

simple_parsing/helpers/partial.py
```diff
@@ -0,0 +1,308 @@
...
+        return target.__name__ + "Config"
+    elif inspect.isfunction(target):
+        return target.__name__ + "_config"
+    raise NotImplementedError(target)
+
+
+class _Partial(_ProtocolMeta):
+    _target_: _C
+
+    def __getitem__(cls, target: type[_T] | Callable[_P, _T]) -> type[Callable[_P, _T]]:
```
JesseFarebro, line 255: Woah, this is wild. So this is actually invoked for just: `Partial[MyClass]`?

JesseFarebro (approved):

simple_parsing/helpers/nested_partial.py
```diff
@@ -0,0 +1,48 @@
+import functools
+from typing import Any, Generic, TypeVar
+
+_T = TypeVar("_T")
+
+
+class npartial(functools.partial, Generic[_T]):
```
JesseFarebro, line 7: Thanks for this! It would still be nice to have instance support in subgroups with frozen dataclasses.

simple_parsing/helpers/partial.py
lebrice: YEah!

examples/partials/partials_example.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+parser = ArgumentParser()
+parser.add_arguments(Config, "config")
+args = parser.parse_args()
+
+
+config: Config = args.config
+print(config)
+expected = "Config(optimizer=AdamConfig(lr=0.0003, beta1=0.9, beta2=0.999, eps=1e-08))"
```
idoby, line 78: Cool! How do you select between SGD and Adam on the command line? An example command would be useful.

simple_parsing/helpers/partial.py
```diff
@@ -0,0 +1,308 @@
...
+
+    IF in some libraries, the signature has a special default value, that we shouldn't use as the
+    default, e.g. "MyLibrary.REQUIRED" or something, then a handler can be registered here to
+    convert it to something else.
+    """
+    return default
+
+
+@overload
+def cache_when_possible(fn: _C) -> _C:
```
idoby, line 69: Personally, I'm not a huge fan of suffixes like `_when_possible` etc. They may be confusing, make discovery harder and are needlessly verbose IMO.
I think a caching mechanism implies that some values will be cached and others won't. The policy should be specified in documentation anyway.
So I'd prefer just `cache`.

simple_parsing/helpers/partial.py
idoby: Personally I'd go for `config_for`, it's short and clear and doesn't get in the way of discoverability (putting in the word dataclass means it will come up when I search for dataclass in my IDE, for no reason)

JesseFarebro:
It would be good to add a test case to make sure `Partial` works with #223. For example,

```py
@dataclasses.dataclass(frozen=True)
class A:
    x: int
    y: bool = True

A1: Type[Partial[A]] = sp.config(A, ignore="x")

@dataclasses.dataclass(frozen=True)
class B:
    a: A = sp.subgroups({
        'a1': A1(y=False),
    })
    
b = sp.parse(B)

assert b(x=1) == A(x=1, y=False)
```

simple_parsing/helpers/partial.py
lebrice: Woops. This is meant to be a module-private function. I'll rename it to `_cache_when_possible`.
And given that it's module-private, I'll will keep things as verbose as needed.

examples/partials/partials_example.py
lebrice: This works same as subgroups, i.e. `--optimizer=sgd --momentum=0.5`.

Check this folder if you want more info: https://github.com/lebrice/SimpleParsing/tree/master/examples/subgroups

idoby: Seems pretty cool

simple_parsing/helpers/partial.py
lebrice: Done

simple_parsing/helpers/nested_partial.py
lebrice: #223

simple_parsing/helpers/partial.py
lebrice: Added a better explanation in here:
https://github.com/lebrice/SimpleParsing/blob/7844d4ed916f2095cecdc921f77213e35da20d2b/simple_parsing/helpers/partial.py#L37-L56

This basically applies a fix that makes it possible to create the Partial classes for the torch optimizer types. 

There's one issue with this though: If pytorch is installed in the environment, it gets loaded here, which can be quite slow.  I'm not 100% sure of the best way to handle this atm.

Do you have any suggestions?

PR #159, opened 2022-08-20
Alias prefix suppression

idoby:
Allow to suppress the prefix defined for a field if used through an alias, for extra-short aliases for often-used command line switches.

### DIFF
--- a/simple_parsing/__init__.py
+++ b/simple_parsing/__init__.py
@@ -5,6 +5,7 @@ from . import helpers, utils, wrappers
 from .conflicts import ConflictResolution
 from .help_formatter import SimpleHelpFormatter
 from .helpers import (
+    Alias,
     MutableField,
     Serializable,
     choice,
@@ -45,6 +46,7 @@ __all__ = [
     "ParsingError",
     "ArgumentGenerationMode",
     "NestedMode",
+    "Alias",
     "InconsistentArgumentError",
 ]
 
--- a/simple_parsing/helpers/fields.py
+++ b/simple_parsing/helpers/fields.py
@@ -32,10 +32,14 @@ K = TypeVar("K")
 V = TypeVar("V")
 T = TypeVar("T")
 
+@dataclasses.dataclass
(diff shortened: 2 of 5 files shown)


idoby:
@lebrice Something seems to be wrong with the 3.6 check. Didn't you drop support for 3.6?

lebrice:
Yeah support was dropped for python 3.6, and I'm not quite sure why it still shows up here...

lebrice:
Hey @idoby, wouldn't passing `field(alias="--out")`be sufficient here?
If not, would you mind clarifying why this is required?

idoby:
> Hey @idoby, wouldn't passing `field(alias="--out")`be sufficient here? If not, would you mind clarifying why this is required?

I'm using prefixes, so specifying aliases still keeps the prefix (e.g. "--experiment_label" becomes "-experiment_l" instead of "-l") and I got tired of typing out the entire thing. So I made a way to drop the prefix for often used cmdline switches.

lebrice:
How would you deal with a dataclass with an "Alias" field being used in two different places?

idoby:
> How would you deal with a dataclass with an "Alias" field being used in two different places?

That would be a conflict. I didn't really get into the details of how AP/SP does conflict resolution, just assumed it would throw an error or something. Do you think this will be a problem?

lebrice:
Wait, @idoby , I'm not sure you answered my question:

wouldn't passing `field(alias="--out")` be sufficient here? (iirc it doesn't use a prefix in that case, right?)

idoby:
> wouldn't passing `field(alias="--out")` be sufficient here? (iirc it doesn't use a prefix in that case, right?)

No, because I'm using a dataclass called ExperimentConfig which is sent to SP with a prefix, namely "experiment".
This DC contains a "label" field with alias "-l", which then becomes "-experiment_l" instead of "-l" like I wanted.

I've included examples in `test_aliases_with_suppressed_prefix()` in `test/test_aliases.py`.

PR #171, opened 2022-10-20
support `dataclasses.InitVar` for python>=3.8

mixilchenko:


test/test_future_annotations.py
```diff
@@ -43,6 +49,9 @@ def test_future_annotations():
     foo = Foo.setup("--a 2 --b heyo --c 1 7.89")
     assert foo == Foo(a=2, b="heyo", c=(1, 7.89))
 
+    foo = Foo.setup("--e 6")
```
mixilchenko, line 52: purest skipif for this test if version < 3.8

lebrice:
Wow, really interesting stuff @mixilchenko ! I'll do a proper review soon!

lebrice (approved):
This is really clean @mixilchenko , thanks for the PR!

I don't doubt that this will come in handy, but just out of curiosity, what use-case did you have in mind while implementing this?

Also, do you think that the init-var arguments should be grouped together on the command-line, or identified differently?
I think it might be a good idea to check how these fields appear on the command-line, and perhaps add a test of some sort so we can verify / keep track of that over the next python releases.

Thanks for doing this, and let me know what you think! :)

mixilchenko:
Speaking about dataclasses, `InitVar` always looked for me like some crutch when 
- you need postprocessing which can't be done without types breaking
- `default` and `default_factory` are not enough to define default value
- you need different types to accept and contain
- ...

I am not so good with examples but here it is (https://mypy-play.net/?mypy=0.981&python=3.10 launchable)
```python
from dataclasses import dataclass, field, InitVar
from functools import cached_property

@dataclass
class A1:
    value: int | str

    def __post_init__(self) -> None:
        self.value = int(self.value)
        
@dataclass
class A2:
    value: int
    
    def __post_init__(self) -> None:
        self.value = int(self.value)
        
@dataclass
class A3:
    arg: InitVar[int | str]
    value: int = field(init=False)
    
    def __post_init__(self, arg: int | str) -> None:
        self.value = int(arg)

class A4:
    arg: int | str
    
    @cached_property
    def value(self) -> int:
        return int(self.arg)

a1 = A1('2')
reveal_type(a1.value)  # Revealed type is "Union[builtins.int, builtins.str]"

a2 = A2('2')  # Argument 1 to "A2" has incompatible type "str"; expected "int"
reveal_type(a2.value)  # Revealed type is "builtins.int"

a3 = A3('2')
reveal_type(a3.value)  # Revealed type is "builtins.int"

a4 = A4('2')
reveal_type(a4.value)  # Revealed type is "builtins.int"
```
In that case, `A3` is preferred over `A4` for me because `arg` is not saved inside


Speaking about argument parser, I don't think that any additional marks are needed. `InitVar`s do look like usual `__init__` arguments and that seems totally reasonable for me.

lebrice (approved):
This looks interesting. I'm kind-of on the edge about how useful this will be, but I see no good reason not to include it. 

Thanks again @mixilchenko !

PR #176, opened 2022-11-01
Add `main` decorator for simple parsing of entry functions

JesseFarebro:
Adds a decorator `main` that will parse command line arguments based on the type annotations of that function.

This decorator supports:
- Positional arguments via Python 3.8+'s support for positional-only arguments with the `/` syntax.
- Nested dataclasses as arguments to a function.
- Automatically parsing the help string based on the docstring of the function.

Example:

```py
import pathlib

import simple_parsing as sp

@sp.decorators.main
def mkdir(path: pathlib.Path, /, *, parents: bool = False, exist_ok: bool = False) -> None:
    """ Make a directory.
    
    Args:
        path: Path at which to create the directory.
        parents: Whether to create parent directories.
        exist_ok: If exist_ok is True `FileExistsError` will be raised.
    """
    path.mkdir(parents=parents, exist_ok=exist_ok)
    
if __name__ == '__main__':
    mkdir()
```

### DIFF
--- a/simple_parsing/__init__.py
+++ b/simple_parsing/__init__.py
@@ -3,6 +3,7 @@
 """
 from . import helpers, utils, wrappers
 from .conflicts import ConflictResolution
+from .decorators import main
 from .help_formatter import SimpleHelpFormatter
 from .helpers import (
     MutableField,
@@ -27,28 +28,29 @@ from .parsing import (
 from .utils import InconsistentArgumentError
 
 __all__ = [
-    "helpers",
-    "utils",
-    "wrappers",
-    "ConflictResolution",
-    "SimpleHelpFormatter",
-    "MutableField",
-    "Serializable",
+    "ArgumentGenerationMode",
+    "ArgumentParser",
     "choice",
+    "ConflictResolution",
+    "DashVariant",
     "field",
     "flag",
+    "helpers",
+    "InconsistentArgumentError",
     "list_field",
+    "main",
     "mutable_field",
+    "MutableField",
+    "NestedMode",
+    "parse_known_args",
+    "parse",
+    "ParsingError",
+    "Serializable",
+    "SimpleHelpFormatter",
     "subgroups",
     "subparsers",
-    "ArgumentParser",
-    "DashVariant",
-    "ParsingError",
-    "parse",
-    "parse_known_args",
-    "ArgumentGenerationMode",
-    "NestedMode",
-    "InconsistentArgumentError",
+    "utils",
+    "wrappers",
 ]
 
 from . import _version
--- a/simple_parsing/decorators.py
+++ b/simple_parsing/decorators.py
@@ -1,0 +1,134 @@
+from __future__ import annotations
+
+import collections
+import dataclasses
+import functools
+import inspect
+import typing
+from typing import Any, Callable, NamedTuple, Optional, Type, TypeVar
+
+import docstring_parser as dp
+from typing_extensions import ParamSpec
+
+from . import helpers, parsing
+
+T = TypeVar("T")
+P = ParamSpec("P")
+
+
+class _Field(NamedTuple):
+    name: str
+    annotation: Type
+    field: dataclasses.Field
+
+
+def _description_from_docstring(docstring: dp.Docstring) -> str:
+    """Construct a description from the short and long description of a docstring."""
+    description = ""
+    if docstring.short_description:
+        description += f"{docstring.short_description}\n"
+        if docstring.blank_after_short_description:
+            description += "\n"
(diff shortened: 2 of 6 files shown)


simple_parsing/decorators.py
```diff
@@ -0,0 +1,122 @@
...
+                default, default_factory = dataclasses.MISSING, dataclasses.MISSING
+                if parameter.default != inspect.Parameter.empty:
+                    if inspect.isfunction(parameter.default):
+                        default_factory = parameter.default
+                    else:
+                        default = parameter.default
+
+                field = _Field(
+                    name,
+                    parameter.annotation,
```
lebrice, line 66: What happens if the user script is using postponed evaluation of type annotations (`from __future__ import annotations`)? I assume that `parameter.annotation` would then be a string. Would this code be affected by that?

The rest of simple-parsing can deal fairly well with postponed evaluations, by evaluating them back into a runtime type annotations (e.g. `"int | bool"` gets brought back down to a `typing.Union[int, bool]` later down the line, you don't have to worry about doing that here)

test/test_decorator.py
```diff
@@ -0,0 +1,117 @@
...
+        ("", 3, partial(_fn_with_all_argument_types, 1, b=1, c=1)),
+        ("2", 4, partial(_fn_with_all_argument_types, b=1, c=1)),
+        ("2 --b=2", 5, partial(_fn_with_all_argument_types, c=1)),
+        ("2 --b=2 --c=2", 6, _fn_with_all_argument_types),
+        ("--c=2", 4, partial(_fn_with_all_argument_types, 1, b=1)),
+        ("--b=2", 4, partial(_fn_with_all_argument_types, 1, c=1)),
+        ("--b=2 --c=2", 5, partial(_fn_with_all_argument_types, 1)),
+    ],
+)
+def test_simple_arguments(
```
lebrice, line 88: These tests are fine for now, but in the future we'd need to make sure this works with other types of fields besides ints.

Just for fun, here's how I parameterize the field types for other tests, which I think is both very hacky and very pretty:

https://github.com/lebrice/SimpleParsing/blob/master/test/test_base.py#L81

Perhaps you could do something similar when constructing your functions (or I could add it myself after this PR).
Note: If I recall correctly, this kind of trickery only works when you type annotations are evaluated at runtime (as you are doing here).

lebrice (approved):
Thanks! Left a few suggestions, but feel free to ignore them if you want.

simple_parsing/__init__.py
```diff
@@ -1,7 +1,7 @@
 """Simple, Elegant Argument parsing.
 @author: Fabrice Normandin
 """
-from . import helpers, utils, wrappers
+from . import decorators, helpers, utils, wrappers
```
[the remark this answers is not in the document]
JesseFarebro: Yeah, I'd prefer this as well.

simple_parsing/decorators.py
```diff
@@ -0,0 +1,122 @@
...
+class _Field(NamedTuple):
+    name: str
+    annotation: Type
+    field: dataclasses.Field
+
+
+def main(original_function: Optional[Callable] = None, **sp_kwargs) -> Callable:
+    """Parse a function's arguments using simple-parsing from type annotations."""
+
+    def _description_from_docstring(docstring: dp.Docstring) -> str:
```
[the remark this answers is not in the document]
JesseFarebro: Good catch!

simple_parsing/decorators.py
```diff
@@ -0,0 +1,122 @@
...
+from . import helpers, parsing
+
+
+class _Field(NamedTuple):
+    name: str
+    annotation: Type
+    field: dataclasses.Field
+
+
+def main(original_function: Optional[Callable] = None, **sp_kwargs) -> Callable:
```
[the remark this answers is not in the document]
JesseFarebro: Yeah, exactly. You can decorate in two ways:

```
@sp.main

@sp.main(**kwargs)
```
where `**kwargs` will be forwarded to `sp.parse`.

test/test_decorator.py
JesseFarebro: The reason I hadn't tested other fields is that those tests should already be covered by SP. We need to check that the decorator constructs the dataclass properly and everything else is handled by `sp.parse`. If we have test coverage for those types in `sp.parse` we should be good.

lebrice: Fair enough.

simple_parsing/decorators.py
lebrice: Re: We can take a look at postponed evaluations later. This should be fine for now.

PR #177, opened 2022-11-11
Fix typing for sp.main decorator

JesseFarebro:


simple_parsing/decorators.py
```diff
@@ -36,17 +32,26 @@ def _description_from_docstring(docstring: dp.Docstring) -> str:
...
-    original_function: Optional[Callable[P, T]] = None,
-    **sp_kwargs,
-) -> Callable[P, T]:
+    original_function: None = None, **sp_kwargs
+) -> Callable[..., Callable[..., Any]]:
+    ...
+
+
+@typing.overload
+def main(original_function: Callable[..., Any], **sp_kwargs) -> Callable[..., Any]:
```
lebrice, line 43: I think that if you use a TypeVar with a bound of Callable instead of Callable[..., Any], then you'll preserve the type signature of the wrapped function.

Something like:
```python

C = TypeVar("C", bound=Callable)

@overload
def main(function: C) -> C:
    ...

@overload
def main(function: None = None) -> Callable[[C], C]:
    ...

def main(function: C | None = None) -> C |Callable[[C], C]:
    ...

```

JesseFarebro: The issue is that if you preserve the signature then you'll get missing parameter warnings when you try to call the function without parameters. e.g.,

```py

@sp.main
def main(a: int):
    ...
    
if __name__ == '__main__':
    main() # Argument missing for parameter `a`.

```

lebrice (approved):

Issue #181, opened 2022-12-06
Future annotations cause add_arguments "NameError" when using Serializable


zhiruiluo:
**Describe the bug**
the ```from __future__ import annotations``` cause add_arguments "NameError" when using Serializable
It will raise the following error:
"NameError: name 'D' is not defined"

**To Reproduce**
```python
# test_simple_parsing.py
from __future__ import annotations
from simple_parsing import Serializable, ArgumentParser, subgroups
from dataclasses import dataclass
import pytest

@dataclass
class MyArguments(Serializable):
    arg1: str = 'this_argment'
    
@pytest.mark.parametrize(
    'sys_argv, result', [
        (['test.py'], 'this_argment'),
        (['test.py', '--arg1', 'test2'], 'test2')
    ],
)
def test_simple_parsing(sys_argv, result):
    parser = ArgumentParser()
    parser.add_arguments(MyArguments, 'myargs')
    args, _ = parser.parse_known_args(sys_argv)
    assert args.myargs.arg1 == result
```

**Expected behavior**

```console
$ pytest test_simple_parsing.py
2 passed
```

**Actual behavior**
A clear and concise description of what is happening.

```console
$ pytest test_simple_parsing.py
2 failed
NameError: name 'D' is not defined
```

**Desktop (please complete the following information):**
 - simple_parsing version: 0.0.21
 - Python version: 3.9.13

**Additional context**

zhiruiluo:
Absuolutely, we will pass the test by removing the '''from __future__ import annotations''' from this file.

zhiruiluo:
Here is my proposed fix #182.
Added an entry to the forward_refs_to_types in get_field_annotation.py to fix the error.
Added a test_issue_181.py file about this issue.

zhiruiluo:
Looks like that this is related to #168.

zhiruiluo:
Thanks for review my code, @lebrice.
The fix is enclosed in #183.
I am closing this issue now.

PR #182, opened 2022-12-06
Modify forward_refs_to_types to fix issue #181

zhiruiluo:
The error was happend in ```get_field_type_from_annotations``` in get_field_annotations.py.

The comments under get_field_type_from_annotations said:
"NOTE: If you get errors of this kind from the function below, then you might want to add an entry to the `forward_refs_to_types` dict above."

I added ```{"D": TypeVar("D")"``` to the forward_refs_to_types. 
If add ```D = TypeVar("D", bound="SerializableMixin")``` to it, this will cause circle import error.

I also added a test_issue_181 for this issue #181 .

All tests have passed with this commit.

### DIFF
--- a/simple_parsing/annotation_utils/get_field_annotations.py
+++ b/simple_parsing/annotation_utils/get_field_annotations.py
@@ -6,7 +6,7 @@ import typing
 from contextlib import contextmanager
 from dataclasses import InitVar
 from logging import getLogger as get_logger
-from typing import Any, Dict, Iterator, Optional, get_type_hints
+from typing import Any, Dict, Iterator, Optional, get_type_hints, TypeVar
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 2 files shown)


simple_parsing/annotation_utils/get_field_annotations.py
```diff
@@ -18,6 +18,7 @@
     "dict": typing.Dict,
     "list": typing.List,
     "type": typing.Type,
+    "D": TypeVar("D"),
```
lebrice, line 21: Hmmm, nice catch, but why exactly is this happening?

lebrice (approved):
I'm not sure why this is happening, but I've gotten reports that this also breaks things in other places, so I'll approve this regardless, and investigate later.

Thanks @zhiruiluo !

Issue #168, opened 2022-10-04
Postponed annotation bug


mixilchenko:
```python
# file a.py
from __future__ import annotations
from dataclasses import dataclass
from pathlib import Path

@dataclass
class A:
    p: Path | None
```
```python
# file b.py
from dataclasses import dataclass
from simple_parsing import ArgumentParser
from a import A

@dataclass
class B(A):
    v: int

if __name__ == '__main__':
    parser = ArgumentParser()
    parser.add_arguments(B, 'b')
    args = parser.parse_args()
    print(args.b)
```

Running `python b.py -h` gives an error
```traceback
Traceback (most recent call last):
  File ".../b.py", line 15, in <module>
    parser.add_arguments(B, 'b')
  File ".../lib/python3.9/site-packages/simple_parsing/parsing.py", line 247, in add_arguments
    new_wrapper = dataclass_wrapper_class(dataclass, dest, prefix=prefix, default=default)
  File ".../lib/python3.9/site-packages/simple_parsing/wrappers/dataclass_wrapper.py", line 60, in __init__
    field_type = get_field_type_from_annotations(self.dataclass, field.name)
  File ".../lib/python3.9/site-packages/simple_parsing/annotation_utils/get_field_annotations.py", line 163, in get_field_type_from_annotations
    annotations_dict = get_type_hints(some_class, localns=local_ns, globalns=global_ns)
  File ".../lib/python3.9/typing.py", line 1458, in get_type_hints
    value = _eval_type(value, base_globals, localns)
  File ".../lib/python3.9/typing.py", line 292, in _eval_type
    return t._evaluate(globalns, localns, recursive_guard)
  File ".../lib/python3.9/typing.py", line 554, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
NameError: name 'Path' is not defined
```

mixilchenko:
Importing `Path` in `b.py` is a workaround

mixilchenko:
I guess [here](https://github.com/lebrice/SimpleParsing/blob/ed8170a32e7765bd98ed42831a428f0cdb645b67/simple_parsing/annotation_utils/get_field_annotations.py#L37-L39) we should also iterate through class mro and add globals from base class modules

lebrice:
Hey @mixilchenko, thanks for posting this.

Yeah that solution makes sense to me.


There are some small complications that might arise, but that we can probably ignore for now.
[...]

PR #183, opened 2022-12-07
Update global namespaces for postponed annotations

zhiruiluo:
This PR is to address the issue #168.
The PR #182 is revoked based on that this fix can also address issue #181.

The bases of dataclasses in a different module from the module of the child dataclass defined were raising ```NameError: name "xxx" is not defined``` for the postponed annotations.

I added a ```get_global_ns``` to recursively get the definition of its bases.
Also, added ```test_postponed_annotation.py``` mentioned in #168.


zhiruiluo:
@lebrice

simple_parsing/annotation_utils/get_field_annotations.py
```diff
@@ -178,9 +177,17 @@ def get_field_type_from_annotations(some_class: type, field_name: str) -> type:
...
+    # Get the globals in the module where the class and the bases of the class were defined, recursively.
+    global_ns = {}
+    
+    def add_global_ns(some_cls: type):
+        nonlocal global_ns
+        global_ns.update(sys.modules[some_cls.__module__].__dict__)
+        for base_cls in some_cls.__bases__:
+            add_global_ns(base_cls)
+
+    add_global_ns(some_class)
```
lebrice, line 189: Correct me if I'm wrong, but I don't think the function, nonlocal, or recursion are needed here:

```suggestion
    global_ns = sys.modules[some_cls.__module__].__dict__
    for base_cls in some_cls.__bases__:
        global_ns.update(sys.modules[base_cls.__module__].__dict__)
```

test/a.py
```diff
@@ -0,0 +1,7 @@
+from __future__ import annotations
```
lebrice, line 1: Would you mind moving these two files one level lower? (e.g. in a `test/postponed_annotations` folder?)

test/test_postponed_annotations.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import annotations
```
lebrice, line 1: To be consistent, this could also be moved one level down, next to the `a.py` and `b.py` files.

test/test_postponed_annotations.py
```diff
@@ -0,0 +1,44 @@
...
+
+@pytest.mark.parametrize(
+    'sys_argv, b_value, p_value', [
+        (['test.py', '--v', '1'], 1, None),
+        (['test.py', '--v', '2', '--p', 'test/'], 2, Path('test/')),
+        (['test.py', '--v', '3', '--p', 'test/test1'], 3, Path('test/test1')),
+        pytest.param(['test.py'], None, None, marks=pytest.mark.xfail(reason='no default value in the dataclass'))
+    ]
+)
+def test_postponed_annotations_with_baseclass(sys_argv, b_value, p_value, monkeypatch):
```
lebrice, line 20: Nitpick: consider typing functions arguments
```suggestion
def test_postponed_annotations_with_baseclass(sys_argv: list[str], b_value: int | None, p_value: Path | None, monkeypatch):
```

test/test_postponed_annotations.py
```diff
@@ -0,0 +1,44 @@
...
+    monkeypatch.setattr(sys, 'argv', sys_argv)
```
lebrice, line 21: Two notes here:
1. You can pass the list of argument arguments to `parse_args` directly. No need to monkeypatch sys.argv here.
2. You might want to use the `TestSetup` class as a base class for class `B`. This way, you could just do
    ```python
    assert B.setup("--v 1 --p test/") == B(v=1, p=Path("test/"))
    ```

lebrice (changes requested):
Thanks @zhiruiluo !

I added some feedback.

simple_parsing/annotation_utils/get_field_annotations.py
zhiruiluo: You are right, @lebrice.
I've double checked the __bases__ attribute. It contains the list of all classes the given class inherits.

test/a.py
zhiruiluo: Done.

simple_parsing/annotation_utils/get_field_annotations.py
zhiruiluo: The recursive get_global_ns will pass the above test.
I am guessing the Fingers in ```FieldComponents("Fingers", "1", 1, "foobar")``` is evaluated afterward and it needs to be evaluated recursively.

We will have to keep ```add_global_ns``` to pass all tests.

zhiruiluo: Let's look at following captured logs for test ```test/test_literal.py::test_literal[verbose-literal_field9]```.
By executing the following code:
```Python
global_ns = sys.modules[some_class.__module__].__dict__
    print(some_class,some_class.__module__)
    for base_cls in some_class.__bases__:
        print(base_cls,base_cls.__module__)
        global_ns.update(sys.modules[base_cls.__module__].__dict__)
```
It failed and prints:
```Python
<class 'test.testutils.test_literal.<locals>.Foo'> test.testutils
<class 'test.testutils.TestSetup'> test.testutils
```

By executing the following code:
```Python
    def add_global_ns(some_cls: type):
        nonlocal global_ns
        global_ns.update(sys.modules[some_cls.__module__].__dict__)
        print(some_cls,some_cls.__module__)
        for base_cls in some_cls.__bases__:
            add_global_ns(base_cls)

    add_global_ns(some_class)
```
It passed and prints:
```Python
<class 'test.test_literal.test_literal.<locals>.Foo'> test.test_literal
<class 'test.testutils.TestSetup'> test.testutils
<class 'object'> builtins
```

The local defined Foo is bounded to a different module in the run time.
That makes us have to evaluate the Foo recursively.
How do you think, @lebrice ?

lebrice: I don't understand the difference. I'll have to look into this.

lebrice: There's also a problem with this implementation, which I mentioned in #168: The namespace of the base classes overwrites the global namespace of the subclass. This is a corner case, and I'm not 100% sure how to solve it.
Naively not overwriting the dictionary entries would cause subtle issues. 
```python
# base.py
from __future__ import annotations
@dataclass
class Foo:
    ...

@dataclass
class Base:
    a: Foo
```
```python
#subclass.py
from __future__ import annotations
from .base import Base
@dataclass
class Foo:
    something_else: bool = True
    
@dataclass
class Subclass(Base):
    other_attribute: Foo
```

With the current code, parsing an instance of `Subclass` would make both attributes have a `base.Foo` attribute, instead of having the right type for each field.

I don't currently have any idea on how to properly solve this, but I'm sure it exists.
You probably don't need to fix this in this PR, but I just wanted you to know about it.

zhiruiluo: It will take me some time to think about that.
I just found some references that might be helpful to read: [postponed annotation in pydantic](https://github.com/pydantic/pydantic/blob/e34ff9255f281eaccac10d0436c082626d9146c3/docs/usage/postponed_annotations.md)
Here is their implementation: [src](https://github.com/pydantic/pydantic/blob/a7656fcd0965e2dfdabbbbe72309dc5211e6115c/pydantic/_internal/_typing_extra.py#L247).

zhiruiluo:
The current commit changed the way of updating global namespaces.
Updating the global_ns from the deepest base until where the field_annotation is defined will address the overwriting issue.
It passed the [Subclass test](https://github.com/lebrice/SimpleParsing/pull/183#discussion_r1043747465) and all others.

simple_parsing/annotation_utils/get_field_annotations.py
```diff
@@ -178,17 +177,22 @@ def get_field_type_from_annotations(some_class: type, field_name: str) -> type:
...
     local_ns: Dict[str, Any] = {"typing": typing, **vars(typing)}
     local_ns.update(forward_refs_to_types)
-    # Get the globals in the module where the class was defined.
-    global_ns = sys.modules[some_class.__module__].__dict__
-
+    # Get the global_ns in the module starting from the deepest base until the module where the field_name is defined.
+    global_ns = {}
+    for base_cls in reversed(some_class.mro()):
+        global_ns.update(sys.modules[base_cls.__module__].__dict__)
+        annotations = getattr(base_cls, '__annotations__', None)
```
lebrice, line 184: I'm surprised, I wouldn't have expected this to work! (since it stops loading the namespaces as soon as it finds a field with a matching name (e.g. as soon as it finds `overwrite_base.Foo`, in the example). :thinking:

test/postponed_annotations/test_postponed_annotations.py
```diff
@@ -0,0 +1,54 @@
...
+def test_postponed_annotations_with_multi_depth_inherits_1():
+    assert P4.setup('--a1 4 --a2 3 --a3 2 --a4 1') == P4(4,3,2,1)
+    
+def test_postponed_annotations_with_multi_depth_inherits_2():
+    assert C.setup('--p test/test1 --v 1 --m string') == C(Path('test/test1'), 1, 'string')
+    
+def test_overwrite_base():
+    import test.postponed_annotations.overwrite_subclass as overwrite_subclass
+    import test.postponed_annotations.overwrite_base as overwrite_base
+    assert overwrite_subclass.Subclass.setup('--something_else False') == overwrite_subclass.Subclass(a=overwrite_base.Foo(), other_attribute=overwrite_subclass.Foo(False))
```
lebrice, line 53: This is really nice! Still surprised that it works though..

lebrice (approved):
Awesome, great work @zhiruiluo ! :smile: 

One last thing. Would you mind applying the pre-commit hooks before merging? (to format, spellcheck, etc)

```console
pip install pre-commit
```

If you want to install and use it in your commits:
```
pre-commit install
```

Otherwise:
```
pre-commit run --all-files
```
Then, ignore any changes to files you haven't touched. In particular, ignore the proposed changes to test_issue_142.py (or something like that).


Worst case, if you don't feel like it, I can apply those later myself.
Thanks again!

zhiruiluo:
> 

Why is the flake8 repo for pre-commit on gitlab? And the link is 404.

simple_parsing/annotation_utils/get_field_annotations.py
zhiruiluo: The loop invariant is that an annotation can only be defined at or above the scope it's used.

zhiruiluo:
Done with the pre-commit check. But, I didn't change the ```.pre-commit-config.yaml```. I am not sure why using gitlab for this, but I did switch to github version of flake8 to continue pre-commit check.

lebrice (approved):
Thanks @zhiruiluo !

Yeah about pre-commit for flake8, it's a bug, I'll fix it, thanks!

Issue #187, opened 2022-12-15
The default_factory for subgroups will not set the default metavar in FieldWrapper


zhiruiluo:
**Describe the bug**
The default_factory for subgroups will not set the default metavar in FieldWrapper.

**To Reproduce**
```python
# test_subgroups_default_factory.py
from typing import Union
from dataclasses import dataclass, is_dataclass, MISSING
import pytest

from simple_parsing import subgroups

from .testutils import TestSetup

@dataclass
class Foo:
    a: int = 1
    b: int = 2

@dataclass
class Bar:
    c: int = 1
    d: int = 2

@dataclass
class Bob(TestSetup):
    thing: Union[Foo, Bar] = subgroups({"foo_thing": Foo, "bar_thing": Bar}, default_factory=lambda: Bar(d=3))
    # the original design will only work when using the "default" argument.
    # thing: Union[Foo, Bar] = subgroups({"foo_thing": Foo, "bar_thing": Bar}, default=Bar(d=3)) 

def test_default_factory_subgroups():
    assert '--thing.c' in Bob.get_help_text('--help')
```

**Expected behavior**
Test passed

**Actual behavior**
Failed 

```console
AssertionError: assert '--thing.c' in ''
```

**Desktop (please complete the following information):**
 - Version: 0.0.21.post1
 - Python version: Python 3.9

**Additional context**
The below ```default_value``` function check both field.default and field.default_factory.
```Python
# utilis.py
def default_value(field: dataclasses.Field) -> T | _MISSING_TYPE:
    """Returns the default value of a field in a dataclass, if available.
    When not available, returns `dataclasses.MISSING`.

    Args:
        field (dataclasses.Field): The dataclasses.Field to get the default value of.

    Returns:
        Union[T, _MISSING_TYPE]: The default value for that field, if present, or None otherwise.
    """
    if field.default is not dataclasses.MISSING:
        return field.default
    elif field.default_factory is not dataclasses.MISSING:  # type: ignore
        constructor = field.default_factory  # type: ignore
        return constructor()
    else:
        return dataclasses.MISSING
```
However, the subgroups give the default value of "default" "None" instead of MISSING.

```Python
def subgroups(
    subgroups: Dict[str, Type[T]],
    *args,
    default: Union[T, Type[T], None] = None,
    **kwargs,
) -> T:
```

zhiruiluo:
Clsed with PR #185.

PR #185, opened 2022-12-13
Rework of the subgroups feature + Minor Refactoring

lebrice:
WIP.

- [x] Add a whole bunch of tests for the subgroups feature (with nesting)
- [x] Make nested subgroups work
- [x] Make other tests work
- [ ] ~~Identify and mark out candidates for subsequent refactoring PRs.~~ (out of scope, this is already a pretty large PR as-is).

PR #188, opened 2022-12-15
Fix subgroups with default_factory

zhiruiluo:
This fix is related to the issue #187.


lebrice:
Hey there @zhiruiluo, thanks for the PR, but I wouldn't touch subgroups too much if I were you. I'm currently rewriting the feature from the ground-up in #185 .

I'll add the changes you mentioned here in there. Thanks!

PR #190, opened 2023-01-06
Fix confusing `default=True` bool behavior (Fix #68)

### LINKED ISSUE #68: Confusing command line options for bool's with `default=True`

Conchylicultor:
I implemented a custom `argparse.Action`. Indeed, this is the **official** solution of argparse added in 3.9: [`argparse.BooleanOptionalAction`](https://docs.python.org/3/library/argparse.html#:~:text=argparse.booleanoptionalaction)

This implementation is inspired from the [official argparse implementation](https://github.com/python/cpython/blob/72263f2a20002ceff443e3a231c713f2e14fe3fe/Lib/argparse.py#L878) with some improvements:

* Support custom `--no` prefix instead of `--no-` (for `absl.flags` compatibility).
* Accept `--flag=true` value
* Support `nargs`
* Support Python 3.8

Mutually exculsive group would be:

* More hacky (flags are parsed independently, so it's difficult to have a global logic)
* Flags would appear either twice int the doc or not be documented
* `argparse.Action` has better error messages, like: `error: the following arguments are required: --my_flag/--nomy_flag`
* Inconsistent with how args are resolved (for example, it is sometimes convenient to be able to overwrite a flag value, for example in bash aliases: `alias my_prog="my_prog --dry_run"`)


simple_parsing/wrappers/field_wrapper.py
```diff
@@ -1040,3 +1028,87 @@ def only_keep_action_args(options: Dict[str, Any], action: Union[str, Any]) -> D
...
+    ):
+        if option_string in self.option_strings:
+            is_neg = option_string.startswith(_BOOL_NO_PREFIX)
+            if values is None:  # --my_flag / --nomy_flag
+                bool_value = not is_neg
+            elif values == []:  # --my_flag / --nomy_flag (with nargs='*')
+                bool_value = [not is_neg]
+            elif is_neg:
+                parser.exit(
+                    message=f"{_BOOL_NO_PREFIX} cannot be used with value (Got: {option_string}={values})"
```
lebrice, line 1102: Just to make sure I totally understand: This means you can set `--verbose=True` but can't set `--noverbose=False`?

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -1040,3 +1028,87 @@ def only_keep_action_args(options: Dict[str, Any], action: Union[str, Any]) -> D
...
+            elif isinstance(values, str):  # --my_flag true
+                bool_value = utils.str2bool(values)
+            elif isinstance(values, list):  # --my_flag true true false
+                bool_value = [utils.str2bool(v) for v in values]
+            else:
+                raise ValueError(f"Unsuported value for {option_string}: {values!r}")
+
+            setattr(namespace, self.dest, bool_value)
+
+    def format_usage(self):
```
lebrice, line 1113: Would this usage str also be correct when using `nargs`?

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -1040,3 +1028,87 @@ def only_keep_action_args(options: Dict[str, Any], action: Union[str, Any]) -> D
...
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: Any,
+        option_string=None,
+    ):
+        if option_string in self.option_strings:
+            is_neg = option_string.startswith(_BOOL_NO_PREFIX)
+            if values is None:  # --my_flag / --nomy_flag
+                bool_value = not is_neg
```
lebrice, line 1097: Shouldn't `self.default` be used at some point? Or is the value ignored? What happens if the flag isn't actually passed?

lebrice (changes requested):
Thanks a lot @Conchylicultor, this is very helpful.

I left a bunch of nitpicky suggestions / comments.
I believe there are a few more things we could explore before merging this.

- How does someone customize the prefix?
It should be easier than setting the module-global variable. It should also ideally be customizable on a per-field basis, rather than one global setting.

I have some suggestions to help with this. For instance, there is already a `flag` utility function for defining a field. Perhaps this function could accept a `negative_prefix="--no"` argument, that would then get stored in the field metadata (the `custom_argparse_args` entry, IIRC).
The `_arg_options` dict is updated with the contents of this dict, so the values would end up getting passed as constructor arguments to your custom Action class later when `add_arguments(**field_wrapper.arg_options)` happens.
There should also be a minimal test to check that this works.

- Will there be backward incompatibility issues? If so, how do we deal with them?
I'm not sure what the right thing to do is here, but it's clear that a small test/example that illustrates the difference would be useful.

Let me know what you think. I know this may seem like a lot, but we can move some of the stuff I mentioned to another PR as well if you want.
 
Again, thank you very much for making this PR. I really appreciate it.

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -1040,3 +1028,87 @@ def only_keep_action_args(options: Dict[str, Any], action: Union[str, Any]) -> D
...
+        default=None,
+        type=None,
+        choices=None,
+        required=False,
+        help=None,
+        metavar=None,
+        nargs=None,
+    ):
+        if nargs is None:
+            nargs = '?'
```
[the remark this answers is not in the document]
lebrice: Oh, one thing might be important to check with this. Do we set a value of `nargs=None` in the _arg_options dict?
If so, we'd need to not set the entry in the first place.

simple_parsing/wrappers/field_wrapper.py
Conchylicultor: `--noflag` is only added when `nargs == '*' '?'` or `nargs >= 1`, so it won't appear in the help otherwise.

test/test_bools.py
```diff
@@ -18,26 +21,81 @@ class Flags(TestSetup):
...
+        ("--a=true", '+', [True]),
+        ("--a true false", '+', [True, False]),
+        # 0 or 1+ argument explicitly required
+        ("--a", '*', [True]),
+        ("--noa", '*', [False]),
+        ("--a=true", '*', [True]),
+        ("--a true false", '*', [True, False]),
+    ],
+)
+def test_bool_nargs(flag, nargs, a):
```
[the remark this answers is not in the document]
Conchylicultor: I find having the `pytest.mark.parametrize` to have the 4 same examples for each `nargs=` makes it more readable easier to think about the difference between each `nargs=`.

Conchylicultor (commented):
* Customizing the `--no` prefix:

Indeed currently this is not possible. Because this seems a quite independent feature. I think it make sense to add it in a separate PR.

* Breaking compatibility:

This will change the default behavior when the default is `True`, which is the current #68 bug this PR fix. With `verbose: bool = True`:

* Before: `--verbose` would set `args.verbose = False` (bug)
* After: `--verbose` set `args.verbose = True`

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -1040,3 +1028,87 @@ def only_keep_action_args(options: Dict[str, Any], action: Union[str, Any]) -> D
...
+    * Support custom prefix (for `absl.flags` compatibility)
+    * Accept `--flag=true` value
+    * Support `nargs`
+    * Support Python 3.8
+
+    """
+
+    def __init__(
+        self,
+        option_strings,
```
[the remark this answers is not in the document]
lebrice: I see what you mean, but I meant having them be annotated directly. 
Also, correct me if I'm wrong, but IDEs wouldn't look for the type hints in the argparse typeshed when using simple-parsing, right?

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -1040,3 +1028,87 @@ def only_keep_action_args(options: Dict[str, Any], action: Union[str, Any]) -> D
...
+        dest,
+        default=None,
+        type=None,
+        choices=None,
+        required=False,
+        help=None,
+        metavar=None,
+        nargs=None,
```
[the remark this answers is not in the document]
lebrice: Sure, then I suggest using `int | Literal["?", "*", "+"] | None`

simple_parsing/wrappers/field_wrapper.py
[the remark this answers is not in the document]
lebrice: Ah I see, agreed.

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -1040,3 +1028,87 @@ def only_keep_action_args(options: Dict[str, Any], action: Union[str, Any]) -> D
...
+    ):
+        if nargs is None:
+            nargs = '?'
+        if not (isinstance(nargs, int) or nargs in {'?', '*', '+'}):
+            raise ValueError(f'nargs={nargs!r} not supported for bools')
+
+        # Do not add `--no` when `nargs >= 1`
+        if nargs == '+' or (isinstance(nargs, int) and nargs >= 1):
+            new_option_strings = option_strings
```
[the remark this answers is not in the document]
lebrice: I don't see the changes though. Did you push your commit? or am I viewing the wrong version?

simple_parsing/wrappers/field_wrapper.py
lebrice: Ok I'll take a look at the tests and resolve this after.

simple_parsing/wrappers/field_wrapper.py
lebrice: I could see myself agreeing with this idea, but as-is, it doesn't seem to be compatible with having nargs="*", right?

I think even though it's confusing, it might be best to keep things uniform for now, and have `--noverbose=False` be the same as `--verbose=True`. We can then discuss this in an issue and turn this off later if needed.

lebrice: Ah nevermind about nargs="*", you only create the negative flag when nargs is for a single argument.
I agree. Let's leave it like this for now then.

simple_parsing/wrappers/field_wrapper.py
lebrice: I see. Makes sense.

test/test_bools.py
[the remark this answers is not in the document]
lebrice: I see your point, but I still consider it important to test the error messages that we get. 
The `match` argument of `pytest.raises` doesn't work perfectly in this case, so you could use `testutils.exits_and_writes_to_stderr`, for instance:
```python
with exits_and_writes_to_stderr("--no cannot be used with value"):
    _ = Foo.setup("--nodebug=True")
```

lebrice (approved):
Hi @Conchylicultor, thanks again for this.

Added some clarifications. I'd appreciate the changes but if you're not up to it, let me know, no worries, I can also add the changes myself in another PR.

Conchylicultor:
Oups, sorry, it looks like VS code didn't pushed my changes.
I'll fix this tomorrow.

(Sorry, I'm not really used to the open source workflow anymore)

simple_parsing/wrappers/field_wrapper.py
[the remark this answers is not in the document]
Conchylicultor: Pushed my changes. Not that `str` is better than `Literal` here because there's some undocumented argparse value for `nargs`. Currently, those will raise error.

simple_parsing/wrappers/field_wrapper.py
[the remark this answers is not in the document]
Conchylicultor: Sorry about that. Pushed the changes

test/test_bools.py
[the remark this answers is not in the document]
Conchylicultor: Updated the tests to check the error message

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -1040,3 +1028,87 @@ def only_keep_action_args(options: Dict[str, Any], action: Union[str, Any]) -> D
...
+            type=type,
+            choices=choices,
+            required=required,
+            help=help,
+            metavar=metavar,
+        )
+
+    def __call__(
+        self,
+        parser: argparse.ArgumentParser,
```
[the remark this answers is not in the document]
Conchylicultor: Done

Conchylicultor:
I pushed my changed. Sorry for the noise.

Let me know what you think

simple_parsing/wrappers/field_wrapper.py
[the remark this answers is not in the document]
lebrice: What value? In which case would it raise an error?

test/test_bools.py
```diff
@@ -18,26 +21,85 @@ class Flags(TestSetup):
...
+        ("--a", '*', []),
+        ("--noa", '*', SystemExit('unrecognized arguments')),
+        ("--a=true", '*', [True]),
+        ("--a true false", '*', [True, False]),
+    ],
+)
+def test_bool_nargs(flag, nargs, a, capsys: pytest.CaptureFixture):
+    @dataclass
+    class MyClass(TestSetup):
+        """Some extension of base-class `Base`"""
```
lebrice, line 96:
```suggestion
        """Some test class"""
```

test/test_bools.py
```diff
@@ -18,26 +21,85 @@ class Flags(TestSetup):
...
+        a: bool = helpers.field(nargs=nargs)
```
lebrice, line 97: By setting `nargs` explicitly, you're bypassing the logic that generates the `nargs` from the annotation and context.
This might be fine, if you're just testing the new Action class on it's own. 
The only thing that bugs me is that the field annotation here is often incorrect (here with nargs="*" you should have a `list[bool]` annotation instead of `bool`).
It could be fine in some situations to manually specify `nargs`, for example if you have a `list[bool]` annotation, you could pass `nargs="+"` or `nargs > 1` to specify the number of values you expect.
Also, FYI, when using `tuple[bool, bool, bool]`, the generated `nargs=3`.

lebrice (approved):
Thanks!

lebrice:
Ok I'm happy with this. Thanks again @Conchylicultor. I'll update the branch and merge it.
I don't think backward-compatibility will be that big of an issue on this. Hopefully this doesn't come back to bite me later!

test/test_bools.py
Conchylicultor: > By setting nargs explicitly, you're bypassing the logic that generates the nargs from the annotation and context.

But this is a limitation of SimpleParsing right ? For example:

```python
@dataclass
class MyClass(TestSetup):
        a: int = helpers.field(nargs='*')
    
flags = MyClass.setup('--a 1 2 3')
assert flags.a == [1, 2, 3]
```

So I don't think this is related to this PR.

Conchylicultor:
Fixed the test

> Does that change make sense?

The `[<>]` is a convention to specify that the `<>` is optional (so `--a [bool]` match `--a` or `--a True`).
Ideally, this should depend on `nargs`:

* `'?'`: `--a [bool]`
* `1`: `--a bool` 
* `'+'`: `--a bool [bool ...]`

But it's likely overengineered and I don't think this should block this PR

test/test_bools.py
lebrice: It's not exactly a limitation of `simple-parsing`. it's just that the annotation you're using (`int`) becomes incorrect if you set `nargs="*"` explicitly.

If you had this, then it would be correct:
```python
@dataclass
class MyClass(TestSetup):
        a: list[int]

flags = MyClass.setup('--a 1 2 3')
assert flags.a == [1, 2, 3]
```

simple_parsing/wrappers/field_wrapper.py
```diff
@@ -1135,7 +1134,11 @@ def __init__(
             choices=choices,
             required=required,
             help=help,
-            metavar=metavar,
+            # TODO: Ideally, this should be depend on `nargs`:
+            # '?': `--a [bool]`
+            # 1: `--a bool` 
+            # '+': `--a bool [bool ...]`
+            metavar=metavar or "[bool]",
```
lebrice, line 1141: Yeah that's a good point, although since this action is only used with plain `bool` fields, then it should probably be fine for now. This action isn't used when you have a `bool | None` field, or `list[bool]`, etc, so the `nargs` would only be different if it were set manually with the `field` function.

I guess we could fix this very minor issue later.

lebrice (approved):

simple_parsing/wrappers/field_wrapper.py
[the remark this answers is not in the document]
Conchylicultor: I'm talking of `argparse.REMAINDER`, `argparse.PARSER`...: https://github.com/python/cpython/blob/cb6ff54b01118e01c26758d2f57cf6da60277bd4/Lib/argparse.py#L101

No idea what those are doing. Currently this will raise an explicit error message in:

`if not (isinstance(nargs, int) or nargs in {'?', '*', '+'}):` would raise an error.

I think this is fine for now. If really required, we could add support for those later.

test/test_bools.py
Conchylicultor: I just meant if we want to fix this behavior (e.g. prevent `nargs='+'` for non-list types), this should be done in another PR as this behavior is not related to this specific action.

simple_parsing/wrappers/field_wrapper.py
Conchylicultor: Actually I think the `[]` should be automatically added here: https://github.com/python/cpython/blob/cb6ff54b01118e01c26758d2f57cf6da60277bd4/Lib/argparse.py#L595

Not sure why this isn't the case. Because this is such a minor issue. I'll be in favor of merging this now and investigate this later.

lebrice:
Hey @Conchylicultor would you mind just correcting the expected strings that are causing the tests to fail?
I'll merge the PR once those pass.

simple_parsing/wrappers/field_wrapper.py
lebrice: Sure, fair enough

test/test_bools.py
lebrice: Fair enough!

Conchylicultor:
Pushed a fix. Tests pass locally for me.

Issue #173, opened 2022-10-24
Subgroups choice isn't serialized in yaml file


kyamagu:
**Describe the bug**
When there is a subgroup, `parse` API fails.

**To Reproduce**
```python
from dataclasses import dataclass

from tempfile import NamedTemporaryFile
from simple_parsing import ArgumentParser, parse, subgroups, Serializable


@dataclass
class ModelConfig(Serializable):
    pass


@dataclass
class ModelAConfig(ModelConfig):
    lr: float = 3e-4
    num_blocks: int = 2


@dataclass
class ModelBConfig(ModelConfig):
    lr: float = 1e-3
    dropout: float = 0.1


@dataclass
class Config(Serializable):
    model: ModelConfig = subgroups(
        {"model_a": ModelAConfig, "model_b": ModelBConfig},
        default=ModelAConfig(),
    )


config = Config()
print(config)
print()

with NamedTemporaryFile(mode="w+", suffix=".yaml") as f:
    config.dump_yaml(f)
    print(open(f.name).read())
    print()
    config2 = parse(Config, config_path=f.name, args=[])
    print(config2)
```

**Expected behavior**
Code should 

```console
Config(model=ModelAConfig(lr=0.0003, num_blocks=2))

model:
  lr: 0.0003
  num_blocks: 2

Config(model=ModelAConfig(lr=0.0003, num_blocks=2))
```

**Actual behavior**
`parse` fails.

```console
Config(model=ModelAConfig(lr=0.0003, num_blocks=2))

model:
  lr: 0.0003
  num_blocks: 2


---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
[<ipython-input-30-a77499e72e1d>](https://localhost:8080/#) in <module>
     35     print(open(f.name).read())
     36     print()
---> 37     config2 = parse(Config, config_path=f.name, args=[])
     38     print(config2)

2 frames
[/usr/local/lib/python3.7/dist-packages/simple_parsing/parsing.py](https://localhost:8080/#) in parse_known_args(self, args, namespace, attempt_to_reorder)
    375                         # A dynamic kind of default value?
    376                         raise RuntimeError(
--> 377                             f"Don't yet know how to detect which subgroup in {subgroup_dict} was "
    378                             f"chosen, if the default value is {value} and the field default is "
    379                             f"{field_wrapper.field.default}"

[...]

PR #197, opened 2023-01-12
Add `replace` function similar to `dataclasses.replace`

zhiruiluo:
The replace function of the dataclasses module has the signature  of [`Dataclasses.replace(obj, /, **changes)`](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace):
- it creates a new object of the same type as `obj`, replacing fields with values from changes.
- If obj is not a Data Class, raises TypeError. If values in changes do not specify fields, raises TypeError.

However, the `Dataclass.replace` doesn't work with nested dataclasses, subgroups, and other features in `simple-parsing`. To solve this, the `simple_parsing.replace` should be supplemented as an extension to `dataclasses.replace`.

The signature of replace function for `simple_parsing` is `simple_parsing.replace(obj: object, changes: Dict[str, Any])` and it supports nested dataclasses, subgroups and other features through `parse` API.

- [x] add replace function
- [x] add test file
- [x] add examples
- [x] add readme


simple_parsing/replace.py
```diff
@@ -0,0 +1,66 @@
...
+        c = OuterClass()
+        c1 = replace(c, changes_1)
+        c2 = replace(c, changes_2)
+        assert c1 == c2
+    """
+
+    _FIELDS = "__dataclass_fields__"
+
+    if not hasattr(type(obj), _FIELDS):
+        raise TypeError("replace() should be called on dataclass instances")
```
lebrice, line 53:
```suggestion
    if not (dataclasses.is_dataclass(obj) and dataclasses.is_dataclass(type(obj)):
        raise TypeError("replace() should be called on dataclass instances")
```

lebrice (changes requested):
Hey there @zhiruiluo , thanks for the PR.

Interesting idea, however I'm not convinced that this formula is the right way to go. I think it would perhaps be better to use the `dict_union`, `to_dict` and `from_dict` utility functions, as I've done here: https://github.com/lebrice/SimpleParsing/blob/master/simple_parsing/helpers/hparams/hyperparameters.py#L200

More importantly, in this context it doesn't make much sense in my opinion for the user to pass values as they would be in the command-line. We don't need to do the round-trip from value -> str -> value again (through the parsing logic). This is also not in-line with the `dataclasses.replace` function, in my opinion.

Let me know what you think.

lebrice:
I'm thinking of something like this:
```python
from simple_parsing.helpers.serialization.serializable import from_dict, to_dict

def replace(dataclass: DataclassT, **new_params) -> DataclassT:
    dataclass_dict = to_dict(dataclass, recurse=True)
    new_dataclass_dict: str = dict_union(dataclass_dict, new_params, recurse=True)
    return from_dict(type(dataclass), new_dataclass_dict, drop_extra_fields=True)
```

zhiruiluo:
> Hey there @zhiruiluo , thanks for the PR.
> 
> Interesting idea, however I'm not convinced that this formula is the right way to go. I think it would perhaps be better to use the `dict_union`, `to_dict` and `from_dict` utility functions, as I've done here: https://github.com/lebrice/SimpleParsing/blob/master/simple_parsing/helpers/hparams/hyperparameters.py#L200
> 
> More importantly, in this context it doesn't make much sense in my opinion for the user to pass values as they would be in the command-line. We don't need to do the round-trip from value -> str -> value again (through the parsing logic). This is also not in-line with the `dataclasses.replace` function, in my opinion.
> 
> Let me know what you think.

My proposed implement still needs more consideration for efficiency for sure, but I am more concerned about the useage cases for the `simple_parsing.replace` directly on dataclasses with dictionary. Let me try your current implementation with my proposed tests first.

zhiruiluo:
> I'm thinking of something like this:
> 
> ```python
> from simple_parsing.helpers.serialization.serializable import from_dict, to_dict
> 
> def replace(dataclass: DataclassT, **new_params) -> DataclassT:
>     dataclass_dict = to_dict(dataclass, recurse=True)
>     new_dataclass_dict = dict_union(dataclass_dict, new_params, recurse=True)
>     return from_dict(type(dataclass), new_dataclass_dict, drop_extra_fields=True)
> ```

Yeah this is much more efficient way. But, the subgroups entries need special care here. For example, 

```python
@dataclass
class A:
    a: float = 0.0

@dataclass
class B:
    b: str = "bar"

@dataclass
class AB:
    a_or_b: A | B = subgroups({"a": A, "b": B}, default="a")

config = AB(a_or_b='a')
# Replacing with {'b': 'test'} is not possible unless
# replacing with {'a_or_b':'b', 'b':'test'}

replace(config,  {'a_or_b':'b', 'b':'test'})

```
Here, the a_or_b has changed its type from A to B where plain dict_union cannot handle this. As per #173, `subgroups` needs additional information to do the serialization as well. That is because of the dynamic strucutre in the nature of `subgroups`.  The run-time replacement of dataclasses is necessary while I am doing grid search of differents models. The master node generates different searching parameter sets for parallel training. Due to the involving of the `subgroups`, which is a really good feature that many people needs, replacement  needs many hand-wire code to achieve.

lebrice:
Hey @zhiruiluo , the `replace` function takes the instantiated subgroup as an input, not a serialized dict. Does that still not work?

zhiruiluo:
That doesn't work.

> Hey @zhiruiluo , the `replace` function takes the instantiated subgroup as an input, not a serialized dict. Does that still not work?

```python
def test_replace_nested_1():
    @dataclass
    class C:
        c: bool = False


    @dataclass
    class D:
        d: int = 0


    @dataclass(frozen=True)
    class CD(TestSetup):
        c_or_d: C | D = subgroups({"c": C, "d": D}, default="c")

        other_arg: str = "bob"

    config = CD.setup('--c_or_d c')
    # passed
    assert replace(config, {"c_or_d": {'c': True}}).c_or_d.c == True
    
    config = CD.setup('--c_or_d d')
    # failed "AttributeError: 'dict' object has no attribute 'd'"
    assert replace(config, {"c_or_d": {'d': 2}}).c_or_d.d == 2
```

zhiruiluo:
Reworked my solution with `to_dict`, `from_dict`, and `dict_union`.
Modifed `to_dict` and `from_dict` to work with `subgroups`.

lebrice (commented):
You're adding an extra entry in the serialized dictionary, and using it in `from_dict`. This makes total sense. 
However this also has the potential to be useful whenever a nested field is of a dataclass type or union of dataclass types, not just  `subgroups`!

For example, In the case where there's a field of type `has_foo: Union[HasFooField, AlsoHasFooField]` where both dataclasses have a field called `foo`, it would be useful to also store the type to use to deserialize the `{"has_foo": {"foo": 123}}` dictionary.

In #201 , I made a side function called [`_nested_field_types`](https://github.com/lebrice/SimpleParsing/pull/201/files#diff-ae3fdb6261d98f818f9df9ac7c6c3f1202591a9fe1d507bf5d26e6d636b75bcdR89) to get all the dataclass types.

In the context of `replace`, I thought it would be best not to touch the `to_dict` and `from_dict` functions.
In your case here, you're only changing the behaviour of these functions if subgroup fields are present, correct? What do you think of instead adding an argument to `from_dict`, `to_dict`, etc that controls whether or not to add this additional information?

test/test_replace.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+@dataclass
+class OuterDataclass(TestSetup):
+    some_arg: str = "some_arg"
+    nested: NestedDataclasses = field(default_factory=NestedDataclasses)
+
+
+@pytest.mark.parametrize(
+    ("config_cls", "args", "changes"),
```
lebrice, line 70: Why does the test use command-line arguments?

test/test_replace.py
```diff
@@ -0,0 +1,239 @@
...
+    logger.info(config_replaced)
+    logger.info(config.setup(args))
+    assert config.setup(args) == config_replaced
+    assert id(config) != id(config_replaced)
+
+
+@pytest.mark.parametrize(
+    ("config_cls", "args", "changes"),
+    [
+        (CD, "--c_or_d d --d 1", {"__subgroups__@c_or_d": "d", "c_or_d.d": 1}),
```
lebrice, line 112: Same here, why does this use command-line arguments?

test/test_replace.py
```diff
@@ -0,0 +1,239 @@
...
+    nested: InnerClass = field(default_factory=InnerClass)
+
+
+def test_replace_outerdataclass():
+    changes_1 = {"outarg": 2, "nested.arg1": 1, "nested.arg2": "bar"}
+    changes_2 = {"outarg": 2, "nested": {"arg1": 1, "arg2": "bar"}}
+    c = OuterClass()
+    c1 = replace(c, changes_1)
+    c2 = replace(c, changes_2)
+    assert c1 == c2
```
lebrice, line 146: Here you could also check that the result is correct:
```suggestion
    assert c1 == c2 == OuterClass(nested=InnerClass(arg1=1, arg2="bar"))
```

test/test_replace.py
```diff
@@ -0,0 +1,239 @@
...
+    class D:
+        d: int = 0
+
+    @dataclass(frozen=True)
+    class CD(TestSetup):
+        c_or_d: C | D = subgroups({"c": C, "d": D}, default="c")
+
+        other_arg: str = "bob"
+
+    config = CD.setup("--c_or_d c")
```
lebrice, line 165: Suggestion: Don't use the command-line in these tests:
```suggestion
    config = CD(c_or_d=C())
```

test/test_replace.py
```diff
@@ -0,0 +1,239 @@
...
+        default_factory=AB,
+    )
+
+
+@pytest.mark.parametrize(
+    ("config_cls", "args", "changes"),
+    [
+        (
+            Config,
+            "--ab_or_cd cd --c_or_d d",
```
lebrice, line 191: Same here. I suggest passing instances of `Config` to the test instead of command-line arguments. This will keep things a bit simpler and more explicit.

test/test_serialization.py
```diff
@@ -0,0 +1,50 @@
...
+class AB(TestSetup, Serializable):
+    integer_only_by_post_init: int = field(init=False)
+    integer_in_string: str = "1"
+    a_or_b: A | B = subgroups({"a": A, "b": B}, default="a")
+
+    def __post_init__(self):
+        self.integer_only_by_post_init = int(self.integer_in_string)
+
+def test_to_dict_from_dict():
+    config = AB.setup("--a_or_b b --b foo --integer_in_string 2")
```
lebrice, line 32: Avoid using the command-line functions here also.

lebrice (changes requested):
Hey @zhiruiluo , I took a better look at your tests. They look good, but there's just one thing I'd like, if possible: to avoid using the command-line functions in the tests, and instead just pass instances of the classes.
I know the tests won't show up as nicely in the output, but that's alright. I'd prefer keeping the tests themselves easier to read and more explicit.

Once that's done, and we've settled on the approach for `to_dict` / `from_dict`, I'll close #201 (which is now just a worse duplicate of yours).

Thanks a lot.

zhiruiluo:
> Hey there @zhiruiluo , hope you're doing good.

Thanks @lebrice! Have my baby girl born recently, so excited!
I am not sure how many types are there, but that's definitely a good starting point we could work on later.

simple_parsing/helpers/serialization/serializable.py
```diff
@@ -652,13 +654,24 @@ def to_dict(dc, dict_factory: type[dict] = dict, recurse: bool = True) -> dict:
...
             continue
 
+        if add_selection:
+            ###### insert subgroups selected key
+            subgroups_dict = f.metadata.get("subgroups")
+            if subgroups_dict:
+                for g_name, g_cls in subgroups_dict.items():
+                    if isinstance(value, g_cls):
+                        _target = f"__subgroups__@{name}"
+                        d[_target] = g_name
```
lebrice, line 664: Nitpick: Reducing the nesting a bit:
```suggestion
        if add_selection and "subgroups" in f.metadata:
            ###### insert subgroups selected key
            subgroups_dict = f.metadata["subgroups"]
            for g_name, g_cls in subgroups_dict.items():
                if isinstance(value, g_cls):
                    _target = f"__subgroups__@{name}"
                    d[_target] = g_name
```

Also, why do we care about the chosen subgroup name here? Can't we just store the class to use?

simple_parsing/helpers/serialization/serializable.py
```diff
@@ -734,8 +752,42 @@ def from_dict(
                 )
             continue
 
-        raw_value = obj_dict.pop(name)
-        field_value = decode_field(field, raw_value, containing_dataclass=cls)
+        if field.metadata.get("subgroups", None) and parse_selection:
+            # decode subgroups from dict
+            subgroups_dict = field.metadata.get("subgroups")
```
lebrice, line 757:
```suggestion
            subgroups_dict = field.metadata["subgroups"]
```

simple_parsing/helpers/serialization/serializable.py
```diff
@@ -734,8 +752,42 @@ def from_dict(
...
+
+            _target = f"__subgroups__@{field.name}"
+            if _target in obj_dict:
+                _target_cls = subgroups_dict[obj_dict[_target]]
+                raw_value = obj_dict.pop(_target)
+
+            if name in obj_dict:
+                raw_value = obj_dict.pop(name)
```
lebrice, line 765: Is it desired for this second case to overwrite the first one if `name in obj_dict` ?
```suggestion
            elif name in obj_dict:
                raw_value = obj_dict.pop(name)
```

simple_parsing/helpers/serialization/serializable.py
lebrice: Ah, we store the subgroup names, because the subgroup class to use would be harder to serialize. That makes sense.

simple_parsing/helpers/serialization/serializable.py
```diff
@@ -792,6 +846,10 @@ def from_dict(
     for name, value in non_init_args.items():
         logger.debug(f"Setting non-init field '{name}' on the instance.")
         setattr(instance, name, value)
+
+    # Call __post_init__ method of the instance
```
lebrice, line 850: Hmmm I don't think this is necessary, since the generated `__init__` or `__new__` from `dataclasses.dataclass` already calls `__post_init__`.

test/test_replace.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+@pytest.mark.parametrize(
+    ("config", "changes"),
+    [
+        (A(a=2.0), {"a": 2.0}),
+        pytest.param(A(), {"b": 1.0}, marks=pytest.mark.xfail(reason="Unrecognized arguments")),
+        (B(b="test"), {"b": "test"}),
+        (AB(a_or_b=A()), {"a_or_b": "a"}),
+        (AB(a_or_b=B(b="foo")), {"__subgroups__@a_or_b": "b", "a_or_b": {"b": "foo"}}),
```
lebrice, line 76: I don't understand this here: Why would we allow these weird keys to be passed as an argument to `replace`?
They aren't fields of the dataclass, so in my opinion they should raise an error.

Feel free to tell me what you think, but I'm thinking that this doesn't look that natural to use. It also mixes up the expected behaviour of `dataclasses.replace` with some command-line-parsing-type logic. I don't think that's something we want `replace` to do.

test/test_replace.py
```diff
@@ -0,0 +1,225 @@
...
+    logger.info(config_replaced)
+    assert config == config_replaced
+    assert id(config) != id(config_replaced)
+
+
+@pytest.mark.parametrize(
+    ("config", "changes"),
+    [
+        (CD(c_or_d=D(d=1)), {"__subgroups__@c_or_d": "d", "c_or_d.d": 1}),
+        (CD(c_or_d=C(c=True)), {"__subgroups__@c_or_d": "c", "c_or_d.c": True}),
```
lebrice, line 106: Not sure that I understand the goal of these. Again, perhaps I'm wrong, feel free to tell me what you think about this one.
Is the idea to allow people to say "Switch the subgroup to 'd' and customize just this value with respect to the default" ?
If so, does this here look a lot better than the equivalent with a more traditional `replace` function?
```python
C_D_subgroups = {"c": C, "d": D}

@dataclass
class Config:
    c_or_d: C | D = subgroups(C_D_subgroups, default="c")

config = Config()
replaced_config = replace(config, c_or_d=C_D_subgroups["d"](d=1))
```

test/test_replace.py
```diff
@@ -0,0 +1,225 @@
...
+        default_factory=AB,
+    )
+
+
+@pytest.mark.parametrize(
+    ("config", "changes"),
+    [
+        (
+            Config(ab_or_cd=CD(c_or_d=D())),
+            {"__subgroups__@ab_or_cd": "cd", "ab_or_cd.__subgroups__@c_or_d": "d"},
```
lebrice, line 184: I have an idea, tell me what you think:
At this point, couldn't the `replace` function take an additional special argument called `subgroups: dict[str, str]` that maps from the field names (flattened) to the subgroup to change?
For instance,  `replace(Config(), subgroups={"ab_or_cd": "cd", "ab_or_cd.c_or_d": "d"})`?
Perhaps it could have a special name like `_subgroups`, if it's possible for the dataclass to have a field called `subgroups`.

We could also not add an argument, but instead use the type of `new_values_dict` to decide if we interpret those values as dataclass fields, or as subgroup choices:
```python
class SubgroupDict(dict):
    ...
replace(Config(), SubgroupDict({"ab_or_cd": "cd", "ab_or_cd.c_or_d": "d"}))
```

Even more interesting, we could use a small wrapper to indicate that a value is the subgroup name, rather than a string value:
```python
class SubgroupName:
    def __init__(self, name: Any):
        self.name = name
replace(Config(), {"ab_or_cd": SubgroupName("cd"), "ab_or_cd.c_or_d": SubgroupName("d"), "ab_or_cd.c_or_d.d": 123})
```

Anyways, it's getting late here, and I'm just rambling at this point. Let me know what you think.

test/test_serialization.py
```diff
@@ -0,0 +1,71 @@
...
+    config_dict = to_dict(config)
+    new_config = from_dict(AB, config_dict, drop_extra_fields=True)
+    assert config == new_config
+
+
+def test_serialization_yaml():
+    config = AB(a_or_b=B(b="foo"), integer_in_string="2")
+    dump_str = config.dumps_yaml()
+    new_config = AB.loads_yaml(dump_str)
+    assert config == new_config
```
lebrice, line 47:
```suggestion
    new_config = AB.loads_yaml(config.dumps_yaml())
    assert new_config == config
    assert new_config.dumps_yaml() == config.dumps_yaml()
```

test/test_serialization.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def test_serialization_json():
+    config = AB(a_or_b=B(b="foo"), integer_in_string="2")
+    dump_str = config.dumps_json()
+    new_config = AB.loads_json(dump_str)
+    assert config == new_config
```
lebrice, line 54:
```suggestion
    new_config = AB.loads_json(config.dumps_json())
    assert new_config == config
    assert new_config.dumps_json() == config.dumps_json()
```

test/test_serialization.py
```diff
@@ -0,0 +1,71 @@
+from __future__ import annotations
```
lebrice, line 1: Just FYI, there is already a `test/utils/test_serialization.py`. I don't mind having more tests, obviously.
This can stay here for now, and I can move things around / merge stuff later if need be.

lebrice (changes requested):
Hi @zhiruiluo , Congratulations!

I added more comments.
I'd like to discuss the `__subgroups__@[key]` feature of `replace` a bit before merging this.

Thanks again!

test/test_serialization.py
zhiruiluo: I've moved to the end of `test/utils/test_serialization.py`.

test/test_replace.py
zhiruiluo: Yeah, you are right. We can do both ways, Wouldn't that be more flexiable for users? I will add your suggestion to the test.

examples/replace/README.md
```diff
@@ -0,0 +1,128 @@
...
+    ab_or_cd: AB | CD = subgroups(
+        {"ab": AB, "cd": CD},
+        default_factory=AB,
+    )
+
+assert Config(ab_or_cd=CD(c_or_d=D())) == sp.replace(
+    {
+        "__subgroups__@ab_or_cd": "cd", 
+        "ab_or_cd.__subgroups__@c_or_d": "d"
+    })
```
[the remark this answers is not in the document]
zhiruiluo: I've modified replace and added a function unflatten_with_selection in replace.py to use get rid of __subgroups__@[key] here.

zhiruiluo:
Hey @lebrice, thanks for your reviewing and feedbacks. They helped a lot.
I agreed some test cases were confusing and `__subgroups__@[key]` makes them less readable. I've tbeen trying to reduce that thing in some cases. However, I don't think we can totally avoid it in few cases.

Here I am giving an example,
```python
@dataclass
class A(TestSetup):
    a: float = 0.0

@dataclass
class B(TestSetup):
    b: str = "bar"

@dataclass
class AB(TestSetup):
    a_or_b: A | B = subgroups({"a": A, "b": B}, default="a")  

def test():
    import unittest
    case = unittest.TestCase()
    case.assertDictEqual(
        to_dict(AB(a_or_b=B(b='test'))), 
        {'__subgroups__@a_or_b': 'b', 'a_or_b': {'b': 'test'}}
    )
```
The `__subgroups__@[key]` has to be appeared in the nested dict to avoid key conflict.
Although, we can still represent it in a plain dictionary that is equilvalent to `AB(a_or_b=B(b='test'))` without using `__subgroups__@[key]` as : `{'a_or_b': 'b', 'a_or_b.b': 'test'}`

test/test_replace.py
zhiruiluo: Those are really good ideas. If I have to pick one from them, I would prefer use the less intrusive way like the small wrapper. But this needs to reconsider for the key conflict problem as I mentioned below.
We could wrapper the `__subgroups__@[key]` in a class that can regularize user's input.

simple_parsing/helpers/serialization/serializable.py
zhiruiluo: It is necessary because the user would override `{"__subgroups__@[key]": "[selected_key]"` generated by `to_dict` by manually putting `{"[key]": "[selected_key]"}` to `replace`

simple_parsing/helpers/serialization/serializable.py
zhiruiluo: This is necessary because the `__post_init__` is after all init.
```python
    try:
        instance = cls(**init_args)  # type: ignore
    except TypeError as e:
        # raise RuntimeError(f"Couldn't instantiate class {cls} using init args {init_args}.")
        raise RuntimeError(
            f"Couldn't instantiate class {cls} using init args {init_args.keys()}: {e}"
        )

    for name, value in non_init_args.items():
        logger.debug(f"Setting non-init field '{name}' on the instance.")
        setattr(instance, name, value)
```
This snippet call `__init__` then, update with the `non_init_args`. This might not agree with the result of `__post_init__`. If I get rid of the calling of `__post_init__`, the test  **test/test_replace.py::test_replace_nested_dict[simple-config2-changes2]** would fail.

test/test_replace.py
zhiruiluo: I mentioned this below.

simple_parsing/helpers/serialization/serializable.py
zhiruiluo: I am little confused now. I checked the `dataclass.replace`. It raise `ValueError` when `init=False`. 
```python
if not f.init:
    # Error if this field is specified in changes.
    if f.name in changes:
        raise ValueError(f'field {f.name} is declared with '
                         'init=False, it cannot be specified with '
                         'replace()')
    continue
```

That contradicts with `from_dict` which allow assign value to field whose `init=False`.

lebrice: Wait, so when we're calling `cls(**init_args)` this **doesn't** call `cls.__post_init__()`? I'm very confused as well!

I think we should probably not set the `init=False` fields in `from_dict` also. But let's remove that in another PR, if that's simpler.

lebrice: I can confirm on my end, with the code of `from_dict` that is currently on the `master` branch:
`cls(**init_args)` **does** call __post_init__:
```python

def test_from_dict_calls_post_init():
    @dataclass
    class Bob:
        a: int = 123
        # Note: Using a default of None so we that if post_init isn't called, we can easily tell.
        negative_a: Optional[int] = field(init=False, default=None)

        def __post_init__(self):
            print(f"Called post-init!")
            self.negative_a = -self.a

    a = 1
    bob = Bob(a=a)
    assert bob.negative_a == -a
    assert to_dict(bob) == {
        "a": a,
        "negative_a": -a,
    }

    constructed_from_dict = from_dict(Bob, {"a": 111})
    assert constructed_from_dict == Bob(a=111)
    assert constructed_from_dict.negative_a == -111
```

As for the non-init fields, leave it as-is, and I'll make another PR to remove this feature after.

simple_parsing/helpers/serialization/serializable.py
```diff
@@ -652,13 +662,23 @@ def to_dict(dc, dict_factory: type[dict] = dict, recurse: bool = True) -> dict:
             d[name] = custom_encoding_fn(value)
             continue
 
+        if add_selection and "subgroups" in f.metadata:
+            ###### insert subgroups selected key
+            subgroups_dict = f.metadata["subgroups"]
+            for g_name, g_cls in subgroups_dict.items():
+                if isinstance(value, g_cls):
```
lebrice, line 669: This check isn't exactly accurate: `subgroups_dict` can contain some `functools.partial` objects as values. What you'd want to use here is `f.metadata["subgroup_dataclass_types"]` : https://github.com/lebrice/SimpleParsing/blob/97ebedb607578f539b5f107120bbd6f479ad8a6f/simple_parsing/helpers/subgroups.py#L100-L102

lebrice (changes requested):
Hi @zhiruiluo, thanks again for doing this and for responding positively to my reviews. 

Here are the contributions of this PR:
1. Adds the ability to replace nested dataclass fields to `dataclasses.replace`, with a new function: `simple_parsing.replace`
2. Fix a bug in the deserialization of dataclasses with subgroups, by also adding the chosen subgroup key to the serialized dictionary.
3. Makes it easier to change the values of subgroup fields, or their attributes, by specifying the
   subgroup key to use, instead of having to pass the value of that subgroup field.

Here is my opinion of each contribution:


## Contribution 1: fixing deserialization issues for dataclasses with subgroup fields

For context, here is the issue:
- When deserializing a dataclass with subgroup fields, the subgroup field is often either not deserialized at all (stays a dictionary), or is deserialized into the first `Serializable` subclass found that has matching fields with the entries of the dictionary. It should instead be deserialized into the correct dataclass type.

I believe there are some issues with your first contribution. Allow me to explain.

1. You're not actually always saving the right value for the chosen subgroup.
You're saving the key for the first dataclass type in the subgroup dict where the value is an instance of that class.
It would therefore be more accurate to actually save the type of the value (e.g. by saving the `__name__` attribute of the value's type), instead of the key of the subgroup dict.

2. I am not in favour of these entries being produced by `to_dict` by default (and expected by `from_dict` by default), at least not until we have a lot more backward-compatibility guarantees.
I'd expect the same exact behaviour as before to occur when using `from_dict` in exactly the same way as before, when using `to_dict` and `from_dict` without passing the `parse_selection`
This means that the bug that can happen where the wrong dataclass is used for deserialization *should still occur in exactly the same way in all cases*!

Once these two are addressed, I'm okay with turning this change on by default, since will only affect the dictionaries produced by dataclasses with subgroup fields, and if it does, it should hopefully only fix deserialization issue: the dictionary for a subgroup field should be deserialized into the right class, instead of just selecting the first class that has matching fields or returning a dictionary.

## Contribution 2: Adding the `simple_parsing.replace` function

I have no issues with this function, as long as it keeps the same scope as what I proposed
previously in #201: it should only ever replace nested fields of the dataclass not subgroups.

## Contribution 3: Replacing subgroups:

I've made up my mind, sorry if it took me a while to decide:
__The `new_values_dict` argument of the `replace` function should only ever contain fields of the dataclass.__

I don't believe that the `"__subgroups__[key]"` syntax for replacing subgroups should be allowed.
The way I see it, there are two ways to rework this feature:

### Option 1: Move it to another function: `replace_subgroups`
```python
from simple_parsing.utils import DataclassT
from simple_parsing.helpers.subgroups import Key

def replace(
    dataclass: DataclassT,
    new_values_dict: dict | None = None,
    **new_values: dict,
) -> DataclassT:
    ...

def replace_subgroups(
    dataclass: DataclassT,
    subgroup_changes: dict[str, Key] | None = None,
) -> DataclassT:
    ...
```
Then, if you wanted to replace the subgroups choice and a value of that subgroup, you'd first use `replace_subgroups`, and then `replace` to replace a value with respect to the subgroup default.

### Option 2: Add another argument to `replace`

If you want to add the `replace_subgroups` functionality in the `replace` function directly (if that's easiser to implement than creating a different function, then it seems to me like it needs to be in a new argument.
This argument would take the **flattened** changes. It could look a little bit like this:

```python
from simple_parsing.utils import DataclassT
from simple_parsing.helpers.subgroups import Key

def replace(
    dataclass: DataclassT,
    new_values_dict: dict | None = None,
    _subgroup_changes: dict[str, Key] | None = None,
    **new_values: dict,
) -> DataclassT:
    ...
```

The dict of subgroup changes here is assumed to be flattened or is able to be flattened without conflicts, otherwise an error is raised. This way, there is no conflict between new values and new subgroup choices.
When compared with having both in the same argument, this also feels a bit more in-line with the `dataclasses.replace` function.


# Conclusion

I suggest that this PR be split into smaller PRs for each contribution. 

For contribution 1), this separate PR would have to:
- reproducee the deserialization issue
- fixes it by saving the type of the subgroup field values in the dict when a new parameter is passed to `to_dict` (turned off by default)
- demonstrates that this change is backward-compatible

For contribution 2) (the "base" `dataclasses.replace` function), we can either re-open and adapt #197 with some of your tests, or slim this current PR down to only include (and test) this contribution.

For contribution 3), it should be a later PR (after the "base" replace function is added). It should also be a separate PR, since it's a new feature, and it should be discussed separately.

I hope this is okay with you. I'm sorry if this is a lot of work, but I think it's important that we get this right, and that we don't break backward-compatibility for existing users.

Please do let me know what you think, and thanks again for all your work on this!

zhiruiluo:
Hey @lebrice, thanks for summarizing the contributios and pointing out a definite plan to improve the whole thing.
I totally agree with your plan and will continue work with you on making this avaiable in the future release.

I will work on the finializing the first contribution as you indicated above. This PR will be suspended as the plan.

lebrice:
Thanks a lot @zhiruiluo , closing this then.

Issue #210, opened 2023-01-24
The deserialization failed on nested dataclasses.


zhiruiluo:
**Describe the bug**

**To Reproduce**
```python
from __future__ import annotations
from simple_parsing import Serializable
from simple_parsing.helpers.serialization import from_dict, to_dict
from dataclasses import dataclass, field
import functools
import pytest

@dataclass
class InnerConfig:
    arg1: int = 1
    arg2: str = 'foo'
    arg1_post_init: str = field(init=False)
    
    def __post_init__(self):
        self.arg1_post_init = str(self.arg1)
    
    
@dataclass
class OuterConfig1(Serializable):
    out_arg: int = 0
    inner: InnerConfig = field(default_factory=InnerConfig)

@dataclass
class OuterConfig2(Serializable):
    out_arg: int = 0
    inner: InnerConfig = field(default_factory=functools.partial(InnerConfig, arg2='bar'))
    
@dataclass
class Level1:
    arg: int = 1
    
@dataclass
class Level2:
    arg: int = 1
    prev: Level1 = field(default_factory=Level1)
    

@dataclass
class Level3:
    arg: int = 1
    prev: Level2 = field(default_factory=Level2)
    

@pytest.mark.parametrize(
    ('config'),
    [
        (OuterConfig1()),
        (OuterConfig2()),
        (Level2()),
        (Level3()),
    ]
)
def test_nested_dataclasses_serialization(config: object):
    config_dict = to_dict(config)
    print(config_dict)
    new_config = from_dict(
        config.__class__,
        config_dict,
        drop_extra_fields=True,
    )
    assert config == new_config
```

**Expected behavior**

```console
All passed
```

**Actual behavior**
A clear and concise description of what is happening.

```console
All failed except `Level2()`
```

**Desktop (please complete the following information):**
 - Version: current master
 - Python version: 3.9

**Additional context**

lebrice:
I'm unable to reproduce the issue:
```
from __future__ import annotations
from dataclasses import dataclass, field
import functools

import pytest
from simple_parsing.helpers.serialization import from_dict, to_dict
from simple_parsing.utils import Dataclass


@dataclass
class Level1:
    arg: int = 1


@dataclass
class Level2:
    arg: int = 1
    prev: Level1 = field(default_factory=Level1)

@pytest.mark.parametrize("config", [Level1(arg=2), Level2(arg=2, prev=Level1(arg=3)),])
def test_nested_dataclass(config: Dataclass):
[...]

PR #211, opened 2023-01-24
Fixing deserialization issues for dataclasses with subgroup fields and nested dataclasses

zhiruiluo:
This PR is the implementation for the contribution 1 mentioned in #197.
- The default values of using add_selected_subgroups and parse_selection were set to `False`. 
- Tests were added to verify the backward-compatiability of `to_dict` and `from_dict`.
- Added fix for issue #210
  - check whether default_factory is dataclass
  - check whether the function of partial of default_factory is dataclass
  - recursively call `from_dict` when either condition is true
- Questions still open to answer
  - whether check field.default for nested dataclass.


lebrice:
Closing in favor of #233

PR #215, opened 2023-02-04
Add `replace_subgroups` function

zhiruiluo:
This PR is to add the `replace_selections` function that replaces some values in a dataclass and replaces dataclass type in nested union of dataclasses or subgroups.

`simple_parsing.replace_selections` function support replacingeplaces dataclass type in nested union of dataclasses or subgroups in addition to `simple_parsing.replace` #212.
- signature: `replace_selections(obj: DataclassT, changes_dict: dict[str, Any] | None = None, selections: dict[str, Key | DataclassT] | None = None, **changes) -> DataclassT:`
  - Compared to `simple_replace.replace`, this calls `replace_selected_dataclass` before calling `simple_parsing.replace`.
- helper function `replace_selected_dataclass`
  - This function replaces the dataclass of subgroups, union, and optional union. The `selections` is in flat format, e.g. {"ab_or_cd": 'cd', "ab_or_cd.c_or_d": 'd'}
   - The values of selections can be `Key` of subgroups, dataclass type, and dataclass instance.

Open question:
- Do we combine the `simple_parsing.replace_selections` and `simple_parsing.replace`?

## Examples
```python
    >>> import dataclasses
    >>> from simple_parsing import replace_selections, subgroups
    >>> from typing import Union
    >>> @dataclasses.dataclass
    ... class A:
    ...     a: int = 0
    >>> @dataclasses.dataclass
    ... class B:
    ...     b: str = "b"
    >>> @dataclasses.dataclass
    ... class Config:
    ...     a_or_b: Union[A, B] = subgroups({'a': A, 'b': B}, default_factory=A)
    ...     a_or_b_union: Union[A, B] = dataclasses.field(default_factory=A)
    ...     a_optional: Union[A, None] = None
    
    >>> base_config = Config(a_or_b=A(a=1))
    # Replace subgroups field by subgroup `Key`, dataclass type, or dataclass instance
    >>> replace_selections(base_config, {"a_or_b.b": "bob"}, {"a_or_b": "b"})
    Config(a_or_b=B(b='bob'), a_or_b_union=A(a=0), a_optional=None)
    >>> replace_selections(base_config, {"a_or_b.b": "bob"}, {"a_or_b": B})
    Config(a_or_b=B(b='bob'), a_or_b_union=A(a=0), a_optional=None)
    >>> replace_selections(base_config, {}, {"a_or_b": B(b="bob")})
    Config(a_or_b=B(b='bob'), a_or_b_union=A(a=0), a_optional=None)
    
    # Replace union of dataclasses and optional dataclass
    >>> replace_selections(base_config, {"a_or_b_union.b": "bob"}, {"a_or_b_union": B})
    Config(a_or_b=A(a=1), a_or_b_union=B(b='bob'), a_optional=None)
    >>> replace_selections(base_config, {"a_optional.a": 10}, {"a_optional": A})
    Config(a_or_b=A(a=1), a_or_b_union=A(a=0), a_optional=A(a=10))
```


zhiruiluo:
WIP

simple_parsing/replace_selections.py
```diff
@@ -0,0 +1,141 @@
...
+def replace_selections(
+    obj: DataclassT,
+    changes_dict: dict[str, Any],
+    selections: dict[str, Key | DataclassT] | None = None,
+) -> DataclassT:
+    ...
+
+
+@overload
+def replace_selections(
```
lebrice, line 36: Do you have a type checker enabled, out of curiosity? (e.g. Pylance, mypy, etc)
You should be getting an error here because `@overload` should be directly above the overloaded function.
Here they are above `replace_selected_dataclass`.

test/test_replace_selections.py
```diff
@@ -0,0 +1,243 @@
...
+        (
+            AllTypes(),
+            {"arg_subgroups.b": "foo", "arg_optional.a": 1.0},
+            {"arg_subgroups": "b", "arg_optional": A},
+            AllTypes(arg_subgroups=B(b="foo"), arg_optional=A(a=1.0)),
+        ),
+    ],
+)
+@pytest.mark.parametrize("pass_dict_as_kwargs", [True, False])
+def test_replace_selections(
```
lebrice, line 173: I think it's important that we test `replace_selected_dataclass` first, by itself.

Also, let's avoid using too many dataclasses in the tests. they are becoming pretty hard to read.

simple_parsing/utils.py
```diff
@@ -938,6 +938,29 @@ def unflatten_split(
...
+) -> PossiblyNestedDict[str, V]:
+    """
+    This function convert flattened dict into additional layer of nested dict
+    and it adds the `keyword` as the key to the unpaired value.
+
+    >>> unflatten_keyword({'ab_or_cd': 'cd', 'ab_or_cd.c_or_d': 'd'})
+    {'ab_or_cd': {'__key__': 'cd', 'c_or_d': 'd'}}
+
+    >>> unflatten_keyword({"a": 1, "b": 2})
+    {'a': {'__key__': 1}, 'b': {'__key__': 2}}
```
lebrice, line 952: I'm not sure I understand the goal of this function. It doesn't seem to be used anywhere besides `replace_selections.py`, so let's keep it in `replace_selections.py` for now.

simple_parsing/replace_selections.py
```diff
@@ -0,0 +1,141 @@
...
+
+def replace_selections(
+    obj: DataclassT,
+    changes_dict: dict[str, Any] | None = None,
+    selections: dict[str, Key | DataclassT] | None = None,
+    **changes,
+) -> DataclassT:
+    """Replace some values in a dataclass and replace dataclass type in nested union of dataclasses or subgroups.
+
+    Compared to `simple_replace.replace`, this calls `replace_selected_dataclass` before calling `simple_parsing.replace`.
```
lebrice, line 105: Can't we just add `replace_selected_dataclass` then?

I'm having trouble understanding all the use-cases for this function.
Correct me if I'm wrong, but you want to be able to replace:
- Subgroup choices with a key
- Subgroup choices with a subgroup value (dataclass type)
- Subgroup choices with a dataclass instance
- Union fields with a value (makes sense)
- Union fields with a dataclass type (?)

I feel like the API is a bit too complicated with all these arguments.
What about having a `replace_subgroup` function, and letting the user compose the two in their own code, just to keep things simpler?
What do you think?

lebrice (changes requested):
Thanks @zhiruiluo, interesting PR.

Left some comments. let me know what you think.

simple_parsing/replace_selections.py
zhiruiluo: I use only Pylance for type checking. I wonder why it didn't report any error for this. May I just overlooked any error or warning for this.

test/test_replace_selections.py
zhiruiluo: Keep only replace_subgroups function in this updating.

simple_parsing/utils.py
zhiruiluo: Sure

simple_parsing/replace_selections.py
zhiruiluo: > Subgroup choices with a key
Subgroup choices with a subgroup value (dataclass type)
Subgroup choices with a dataclass instance
Union fields with a value (makes sense)
Union fields with a dataclass type (?)

Yes you are right.
Union fields with dataclass type will be initialized working like a default factory.

zhiruiluo:
Hi @lebrice,
Sorry for letting you wait too long.

Here is the updating:
- keep only replace_subgroups function
- simplify test cases
- move replace_subgroups function into replace.py file
- move the unflatten function out of utils.py

simple_parsing/replace.py
```diff
@@ -90,3 +104,80 @@ def replace(obj: DataclassT, changes_dict: dict[str, Any] | None = None, **chang
...
+) -> PossiblyNestedDict[str, V]:
+    """
+    This function convert a flattened dict into a nested dict
+    and it inserts the `keyword` as the selection into the nested dict.
+
+    >>> unflatten_selection_dict({'ab_or_cd': 'cd', 'ab_or_cd.c_or_d': 'd'})
+    {'ab_or_cd': {'__key__': 'cd', 'c_or_d': 'd'}}
+
+    >>> unflatten_selection_dict({"a": 1, "b": 2})
+    {'a': {'__key__': 1}, 'b': {'__key__': 2}}
```
lebrice, line 120: I don't understand this case here. There isn't any overlap between the keys, so shouldn't it return the same dict unchanged?

simple_parsing/replace.py
```diff
@@ -90,3 +104,80 @@ def replace(obj: DataclassT, changes_dict: dict[str, Any] | None = None, **chang
...
+    The `selections` dict can be in flat format or in nested format.
+
+    The values of selections can be `Key` of subgroups, dataclass type, and dataclass instance.
+    """
+    keyword = "__key__"
+
+    if selections:
+        selections = unflatten_selection_dict(selections, keyword)
+    else:
+        return obj
```
lebrice, line 143:
```suggestion
    if not selections:
        return obj
    selections = unflatten_selection_dict(selections, keyword)
```

simple_parsing/replace.py
```diff
@@ -90,3 +104,80 @@ def replace(obj: DataclassT, changes_dict: dict[str, Any] | None = None, **chang
...
+                field_value = field.default_factory()
+            else:
+                raise ValueError(f"invalid selection key '{key}' for field '{field.name}'")
+
+            if child_selections:
+                new_value = replace_subgroups(field_value, child_selections)
+            else:
+                new_value = field_value
+        else:
+            raise ValueError(f"The replaced subgroups contains no dataclass in its annotation {t}")
```
lebrice, line 177:
```suggestion
        if not contains_dataclass_type_arg(t):
            raise ValueError(f"The replaced subgroups contains no dataclass in its annotation {t}")
        child_selections = selections.pop(field.name)
        key = child_selections.pop(keyword, None)

        if is_dataclass_type(key):
            field_value = key()
        elif is_dataclass_instance(key):
            field_value = copy.deepcopy(key)
        elif field.metadata.get("subgroups", None):
            field_value = field.metadata["subgroups"][key]()
        elif is_optional(t) and key is None:
            field_value = None
        elif contains_dataclass_type_arg(t) and key is None:
            field_value = field.default_factory()
        else:
            raise ValueError(f"invalid selection key '{key}' for field '{field.name}'")

        if child_selections:
            new_value = replace_subgroups(field_value, child_selections)
        else:
            new_value = field_value
```

simple_parsing/replace.py
```diff
@@ -90,3 +104,80 @@ def replace(obj: DataclassT, changes_dict: dict[str, Any] | None = None, **chang
...
+        if contains_dataclass_type_arg(t):
+            child_selections = selections.pop(field.name)
+            key = child_selections.pop(keyword, None)
+
+            if is_dataclass_type(key):
+                field_value = key()
+            elif is_dataclass_instance(key):
+                field_value = copy.deepcopy(key)
+            elif field.metadata.get("subgroups", None):
+                field_value = field.metadata["subgroups"][key]()
```
lebrice, line 164: Here you're assuming that the subgroup values are callables, right?
They could also be frozen dataclass instances (since #223 )

simple_parsing/replace.py
```diff
@@ -90,3 +104,80 @@ def replace(obj: DataclassT, changes_dict: dict[str, Any] | None = None, **chang
...
+
+            if child_selections:
+                new_value = replace_subgroups(field_value, child_selections)
+            else:
+                new_value = field_value
+        else:
+            raise ValueError(f"The replaced subgroups contains no dataclass in its annotation {t}")
+
+        if not field.init:
+            raise ValueError(f"Cannot replace value of non-init field {field.name}.")
```
lebrice, line 180: I suggest moving this before all the code above.

lebrice (changes requested):
Thanks @zhiruiluo , this is a lot clearer!

There are just a few small things to tweak, but this looks pretty good!

simple_parsing/replace.py
zhiruiluo: Will fix this.

simple_parsing/replace.py
zhiruiluo: Will check is_dataclass_instance for the selection to distinguish fronzen dataclass obj or dataclass type.

simple_parsing/replace.py
zhiruiluo: Fixed.

zhiruiluo: Fixed the unflatten_selection_dict and add recursive flag to it. It passed those tests:
```python
    >>> unflatten_selection_dict({'ab_or_cd': 'cd', 'ab_or_cd.c_or_d': 'd'})
    {'ab_or_cd': {'__key__': 'cd', 'c_or_d': 'd'}}
    
    >>> unflatten_selection_dict({'lv1': 'a', 'lv1.lv2': 'b', 'lv1.lv2.lv3': 'c'})
    {'lv1': {'__key__': 'a', 'lv2': 'b', 'lv2.lv3': 'c'}}

    >>> unflatten_selection_dict({'lv1': 'a', 'lv1.lv2': 'b', 'lv1.lv2.lv3': 'c'}, recursive=True)
    {'lv1': {'__key__': 'a', 'lv2': {'__key__': 'b', 'lv3': 'c'}}}

    >>> unflatten_selection_dict({'ab_or_cd.c_or_d': 'd'})
    {'ab_or_cd': {'c_or_d': 'd'}}

    >>> unflatten_selection_dict({"a": 1, "b": 2})
    {'a': 1, 'b': 2}
```

zhiruiluo:
Hi @lebrice, Here are some updates so far:
- Fixed the bug in unflatten_selection_dict when no overlapping of keys and added a recursive flag.
- reordered some logics for readability
- add test case for the frozen dataclass to work

I would appreciate any further feedback or suggestions for improvement that you may have. Thank you.

simple_parsing/replace.py
```diff
@@ -90,3 +105,125 @@ def replace(obj: DataclassT, changes_dict: dict[str, Any] | None = None, **chang
...
+) -> PossiblyNestedDict[str, V]:
+    """
+    This function convert a flattened dict into a nested dict
+    and it inserts the `keyword` as the selection into the nested dict.
+
+    >>> unflatten_selection_dict({'ab_or_cd': 'cd', 'ab_or_cd.c_or_d': 'd'})
+    {'ab_or_cd': {'__key__': 'cd', 'c_or_d': 'd'}}
+    
+    >>> unflatten_selection_dict({'lv1': 'a', 'lv1.lv2': 'b', 'lv1.lv2.lv3': 'c'})
+    {'lv1': {'__key__': 'a', 'lv2': 'b', 'lv2.lv3': 'c'}}
```
lebrice, line 121: I would suggest making `recursive=True` by default. The non-recursive case seems a bit strange to me.
Also, is this function expected to be used by users? If not, I'd prefix it with an underscore (`_unflatten_selection_dict`) to signal that it is an internal function only used in this module.
I would also place it below other "public" functions like `replace` and `replace_subgroups`.

lebrice (approved):
Thanks @zhiruiluo , and sorry it took so long to review this.

This looks good to me. I only have a tiny comment on the naming of one of the functions, but appart from that, this is good to merge!

Thanks again!

simple_parsing/replace.py
zhiruiluo: Thanks @lebrice, I have made changes for all your suggestions.
But the test failed due to the Literal issue #259. Once #260 merged, this PR will pass the test.

lebrice (approved):

PR #221, opened 2023-02-27
Literal types no longer cause a warning on decoding

norabelrose:
On `master`, you can't deserialize a dataclass with `Literal` type annotations without running into warnings like this:
```
UserWarning: Unable to find a decoding function for the annotation typing.Literal['first', 'last', 'mean'] (of type <class 'typing._LiteralGenericAlias'>). Will try to use the type as a constructor. Consider registering a decoding function using `register_decoding_fn`, or posting an issue on GitHub.
```
I've fixed this by adding an extra if-block in `helpers/serialization/decoders.py` that checks for Literal types:
```
if is_literal(t):
        logger.debug(f"Decoding a Literal field: {t}")
        possible_vals = get_type_arguments(t)
        return decode_literal(*possible_vals)
```
The Callables created by the `decode_literal` function simply check if the input values are among those permitted by the Literal type, and raise a `TypeError` if so.

Please let me know if there's anything else I need to do (e.g. add a test?) to make this PR mergeable.


lebrice:
Hey there @norabelrose, thanks for making this!

You're right that this would ideally require a test. I'll get back to you with an example of what that could look like, but in the meantime, you could probably take a look at this:
- https://github.com/lebrice/SimpleParsing/blob/master/test/test_literal.py#L54-L67
- https://github.com/lebrice/SimpleParsing/blob/master/test/test_decoding.py

The goal of the tests would be to:
1. reproduce that there is indeed raising a warning without the fix
2. that the fix removes it
3. check that the decoding works as expected.

If you feel like taking a swing at this, let me know. If not, no worries! :)
Let me know what you think.

simple_parsing/helpers/serialization/decoding.py
```diff
@@ -11,7 +11,7 @@
 from functools import lru_cache, partial
 from logging import getLogger
 from pathlib import Path
-from typing import Any, Callable, TypeVar
+from typing import Any, Callable, Literal, TypeVar
```
lebrice, line 14: This is what's causing an issue: It's not in `typing` yet for Python 3.7.
```suggestion
from typing import Any, Callable, TypeVar
from typing_extensions import Literal
```

lebrice (approved):

norabelrose:
All right let me take a look at the tests

norabelrose:
Hmmm I'm actually not able to replicate the bug anymore? Not sure what's going on

norabelrose:
Nvm now I'm seeing the bug again

norabelrose:
Just added a test; should fix #226

test/test_decoding.py
```diff
@@ -33,6 +34,24 @@ class SomeClass(Serializable):
...
+def test_literal_decoding():
+    @dataclass
+    class SomeClass(Serializable):
+        x: Literal["a", "b", "c"] = "a"
+
+    # This test should fail if there's a warning on decoding- previous versions
+    # have raised a UserWarning when decoding a literal, of the form:
+    # Unable to find a decoding function for annotation typing.Literal['a', 'b', 'c']
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
```
lebrice, line 46: I guess the test to replicate the issue would have used `pytest.warns`, right?
Since we expect the test not to warn anymore, I think it's fine to make this a comment.

lebrice (approved):
Thanks @norabelrose !

test/test_decoding.py
norabelrose: Ok! Feel free to do that yourself, or I can do it in a bit

test/test_decoding.py
```diff
@@ -1,9 +1,10 @@
 import json
 from dataclasses import dataclass
 from test.testutils import Generic, TypeVar
-from typing import Any, Dict, List, Optional, Tuple, Type
+from typing import Any, Dict, List, Literal, Optional, Tuple, Type
```
lebrice, line 4:
```suggestion
from typing import Any, Dict, List, Optional, Tuple, Type
from typing_extensions import Literal
```

test/test_decoding.py
```diff
@@ -33,6 +34,24 @@ class SomeClass(Serializable):
...
+    class SomeClass(Serializable):
+        x: Literal["a", "b", "c"] = "a"
+
+    # This test should fail if there's a warning on decoding- previous versions
+    # have raised a UserWarning when decoding a literal, of the form:
+    # Unable to find a decoding function for annotation typing.Literal['a', 'b', 'c']
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+
+        assert SomeClass.loads('{"x": "a"}') == SomeClass()
```
lebrice, line 48:
```suggestion
    # with pytest.warns(UserWarning, match="Unable to find a decoding function"):
    #     assert SomeClass.loads('{"x": "a"}') == SomeClass()
```

lebrice (approved):

PR #228, opened 2023-03-02
fix postponed annotation when field is overwritten

mixilchenko:
Thanks to @zhiruiluo #183 covers the main problem described in #168.
However, I want to cover one more case when subclass definition overwrites variable type with specific subtype.

### DIFF
--- a/simple_parsing/annotation_utils/get_field_annotations.py
+++ b/simple_parsing/annotation_utils/get_field_annotations.py
@@ -5,6 +5,7 @@ import types
 import typing
 from contextlib import contextmanager
 from dataclasses import InitVar
+from itertools import dropwhile
 from logging import getLogger as get_logger
 from typing import Any, Dict, Iterator, Optional, get_type_hints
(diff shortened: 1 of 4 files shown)


lebrice:
Hey @mixilchenko , thanks for making this.

I'm having a bit of trouble understanding what case this is fixing. Would you mind adding some docstrings in the test explaining a bit more what the test is checking? It's not thaaat easy to understand the test from just looking at the current code.

lebrice (changes requested):
Hey @mixilchenko , thanks for making this.

I'm having a bit of trouble understanding what case this is fixing. Would you mind adding some docstrings in the test explaining a bit more what the test is checking? It's not thaaat easy to understand the test from just looking at the current code.

mixilchenko:
I tried to improve naming and added some docstrings to highlight the difference between these two tests

lebrice (approved):
Thanks @mixilchenko !

PR #248, opened 2023-04-19
Fix #247

### LINKED ISSUE #247: Setting `add_config_path_arg=True` in `ArgumentParser` overwrites options with defaults
**Describe the bug**
Creating an `ArgumentParser` sets the values of
```
        FieldWrapper.add_dash_variants = add_option_string_dash_variants
        FieldWrapper.argument_generation_mode = argument_generation_mode
        FieldWrapper.nested_mode = nested_mode
```

so creating an `ArgumentParser` with default arguments inside parsing (as happens with `add_config_path_arg=True`) overwrites any custom settings of these.

**To Reproduce**

Create the program `reproduce.py`:

```python
import simple_parsing


@dataclass
class Baz:
    arg: int = 2


@dataclass
class Foo:
    baz: Baz = Baz()


if __name__ == "__main__":
    parser = simple_parsing.ArgumentParser(
        add_config_path_arg=True,  # setting to False works fine
        argument_generation_mode=simple_parsing.ArgumentGenerationMode.NESTED,
    )
    parser.add_arguments(Foo, "foo")
    args = parser.parse_args()
```
and run `python reproduce.py --help`

**Expected behavior**
At the very bottom, the argument `arg` should be accessible via `--foo.baz.arg`. It should print this:

```console
usage: reproduce.py [-h] [--foo.baz.arg int]

options:
  -h, --help         show this help message and exit

Foo ['foo']:
  Foo(baz: __main__.Baz = Baz(arg=2))

Baz ['foo.baz']:
  Baz(arg: int = 2)

  --foo.baz.arg int  (default: 2)
```

**Actual behavior**
Prints

```console
usage: reproduce.py [-h] [--config_path Path] [--arg int]

options:
  -h, --help          show this help message and exit
  --config_path Path  Path to a config file containing default values to use. (default: None)

Foo ['foo']:
  Foo(baz: __main__.Baz = Baz(arg=2))

Baz ['foo.baz']:
  Baz(arg: int = 2)

  --arg int           (default: 2)
```

rhaps0dy:
Pass the `FieldWrapper.` globals when an `ArgumentParser` is created again.

More bugs of this kind can emerge in the future, but this is the minimal change to fix this.

### DIFF
--- a/simple_parsing/parsing.py
+++ b/simple_parsing/parsing.py
@@ -298,7 +298,13 @@ class ArgumentParser(argparse.ArgumentParser):
                 self.set_defaults(config_file)
 
         if self.add_config_path_arg:
-            temp_parser = ArgumentParser(add_config_path_arg=False, add_help=False)
+            temp_parser = ArgumentParser(
+                add_config_path_arg=False,
+                add_help=False,
+                add_option_string_dash_variants=FieldWrapper.add_dash_variants,
+                argument_generation_mode=FieldWrapper.argument_generation_mode,
+                nested_mode=FieldWrapper.nested_mode,
+            )
             temp_parser.add_argument(
                 "--config_path",
                 type=Path,


lebrice:
Awesome, thanks a lot @rhaps0dy !

PR #256, opened 2023-04-28
added example for how to use the flag helper

LTluttmann:
hey, I just discovered this awesome library. Since I needed it and could find an example for it, I made a simple example for how to use the flag() field. Hope this is useful.


lebrice (approved):
Thanks a lot @LTluttmann !

PR #258, opened 2023-05-04
Serialized none vals

### LINKED ISSUE #257: Allow for None values in serialized namespaces
**Is your feature request related to a problem? Please describe.**
Currently, when reading a serialized dataclass which contains None values, an error occurs during decoding the serialized data. 

To reproduce 

```
from simple_parsing.helpers import Serializable
import tempfile

@dataclass
class A(Serializable):
    a: float = None

a = A()

with tempfile.NamedTemporaryFile("w+", suffix=".json") as fp:
    a.save(fp.name)
    b = A.load(fp.name)

```
This raises the following error: "float() argument must be a string or a real number, not 'NoneType'". 

**Describe the solution you'd like**
The error is raised in the decode_field function, [here](https://github.com/lebrice/SimpleParsing/blob/ce027d22cc843177dfb317251165690ea4fae4b6/simple_parsing/helpers/serialization/serializable.py#L813). One solution would be to insert an if-else statement like so:

```
if raw_value is None:
    field_value = None
else:
    field_value = decode_field(
        field, raw_value, containing_dataclass=cls, drop_extra_fields=drop_extra_fields
    )

```

However, I don't know if this might introduce problems elsewhere or with other edge cases. Tests are still working, though

**Describe alternatives you've considered**
Maybe, there is also a workaround I'm not aware of

LTluttmann:
Fixes #257

### DIFF
--- a/simple_parsing/helpers/serialization/serializable.py
+++ b/simple_parsing/helpers/serialization/serializable.py
@@ -810,9 +810,12 @@ def from_dict(
             continue
 
         raw_value = obj_dict.pop(name)
-        field_value = decode_field(
-            field, raw_value, containing_dataclass=cls, drop_extra_fields=drop_extra_fields
-        )
+        if raw_value is None:
+            field_value = None
+        else:
+            field_value = decode_field(
+                field, raw_value, containing_dataclass=cls, drop_extra_fields=drop_extra_fields
+            )
 
         if field.init:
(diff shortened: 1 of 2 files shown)


lebrice (changes requested):
Hi @LTluttmann, I'm not convinced that this is a good idea. (Let's chat on #257 )

Issue #259, opened 2023-05-29
Literals not parsing correctly in python 3.9


levmckinney:
**Describe the bug**
`Literals` appear to be broken in python 3.9. Valid inputs are not recognized as valid during parsing. Interestingly, I do not appear to have this issue in python 3.10

**To Reproduce**
```
from typing import Literal
from dataclasses import dataclass
from simple_parsing import ArgumentParser

@dataclass
class Foo:
    param: Literal['bar', 'biz'] = 'biz'


def test_parsing():
    parser = ArgumentParser()
    parser.add_arguments(Foo, dest="foo")
    args = parser.parse_args()
    print(args.foo.param)

if __name__ == '__main__':
    test_parsing()
```

**Expected behavior**

```console
$ python issue.py
biz
```

**Actual behavior**
A clear and concise description of what is happening.

```console
$ python issue.py
usage: issue.py [-h] [--param typing.Literal['bar', 'biz']]
issue.py: error: argument --param: invalid typing.Literal['bar', 'biz'] value: 'biz'
$ python issue.py --param biz
usage: issue.py [-h] [--param typing.Literal['bar', 'biz']]
issue.py: error: argument --param: invalid typing.Literal['bar', 'biz'] value: 'biz'
```

**Desktop (please complete the following information):**
 - Version: simple-parsing==0.1.2.post1
 - Python version: 3.9

**Additional context**
I am not encountering this bug in python 3.10 it only appears to be causing my python 3.9 CI to fail.

Complete pip freeze
```
accelerate==0.19.0
aiohttp==3.8.4
aiosignal==1.3.1
appdirs==1.4.4
asttokens==2.2.1
async-timeout==4.0.2
attrs==23.1.0
backcall==0.2.0
certifi==2023.5.7
charset-normalizer==3.1.0
click==8.1.3
cmake==3.26.3
comm==0.1.3
coverage==7.2.6
datasets==2.12.0
debugpy==1.6.7
decorator==5.1.1
dill==0.3.6
docker-pycreds==0.4.0
docstring-parser==0.15
exceptiongroup==1.1.1
executing==1.2.0
fastjsonschema==2.17.1
filelock==3.12.0
frozenlist==1.3.3
fsspec==2023.5.0
gitdb==4.0.10
GitPython==3.1.31
huggingface-hub==0.14.1
idna==3.4
importlib-metadata==6.6.0
iniconfig==2.0.0
ipykernel==6.23.1
ipython==8.13.2
[...]

PR #260, opened 2023-06-01
Fix literals

brentyi:
Here's a fix for #259!

Could use tests.


lebrice (approved):
Thanks @brentyi !

lebrice:
Fixes #259

Issue #269, opened 2023-07-01
test_decorator failure for Python 3.11.4


zhiruiluo:
**Describe the bug**
test_decorator.py failed during collection for Python 3.11.4 at line 65.
```python
>> signature = signature.replace(parameters=new_parameters)  
```
```bash
test/test_decorator.py:86: in <module>
    ("--b=2", 4, partial(_fn_with_all_argument_types, 1, c=1)),
test/test_decorator.py:66: in partial
    signature = signature.replace(parameters=new_parameters)
conda/envs/py311test/lib/python3.11/inspect.py:3052: in replace
    return type(self)(parameters,
.conda/envs/py311test/lib/python3.11/inspect.py:3008: in __init__
    raise ValueError(msg)
E   ValueError: non-default argument follows default argument
---------------------------------------------------------------
```
The test case ```("--b=2", 4, partial(_fn_with_all_argument_types, 1, c=1)),``` has parameters ```[<Parameter "a: int = 1">, <Parameter "b: int">, <Parameter "c: int = 1">]```

**To Reproduce**
```
pytest test/test_decorator.py
```

**Desktop (please complete the following information):**
 - Version: master
 - Python version: 3.11.4

**Additional context**

PR #270, opened 2023-07-01
Fix test_decorator issue #269

zhiruiluo:
To fix issue #269.

### DIFF
--- a/test/test_decorator.py
+++ b/test/test_decorator.py
@@ -44,27 +44,8 @@ class ListWithPopDefault(collections.UserList):
 
 
 def partial(fn: Callable, *args, **kwargs) -> Callable:
-    """Partial via changing the signature defaults."""
-
-    @functools.wraps(fn)
-    def _wrapper(*other_args, **other_kwargs):
-        return fn(*other_args, **other_kwargs)
-
-    signature = inspect.signature(fn)
-    parameters = signature.parameters
-
-    args = ListWithPopDefault(args)
-
-    new_parameters = []
-    for key, param in parameters.items():
-        param = typing.cast(inspect.Parameter, param)
-        default = args.pop(0, None) or kwargs.get(key, None)
-        if default is not None:
-            param = param.replace(default=default)
-        new_parameters.append(param)
-
-    signature = signature.replace(parameters=new_parameters)
-    _wrapper.__signature__ = signature
+    _wrapper = functools.partial(fn, *args, **kwargs)
+    _wrapper.__qualname__ = fn.__qualname__
 
     return _wrapper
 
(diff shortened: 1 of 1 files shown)


zhiruiluo:
This doesn't exactly behave like the original partial function via changing defaults.

zhiruiluo:
Added a ```delay_evaluation_wrapper``` to prevent collection error.

test/test_decorator.py
```diff
@@ -69,29 +75,41 @@ def _wrapper(*other_args, **other_kwargs):
...
+
+
 @pytest.mark.parametrize(
-    "args, expected, fn",
+    "args, expected, delay_wrapper",
     [
         ("", 1, partial(_fn_with_positional_only, 1)),
         ("2", 2, partial(_fn_with_positional_only, 1)),
-        ("2", 2, _fn_with_positional_only),
+        ("2", 2, partial(_fn_with_positional_only)),
```
lebrice, line 94: Hey there @zhiruiluo , Would you mind explaining a few things:
- why is the `delay_wrapper` is necessary here?
- Did the `inspect` module change from 3.11 to 3.11.3? Is that what broke this test?

zhiruiluo: The recently PR https://github.com/python/cpython/pull/103557 disallows pos-or-kw params without default after pos-only with default, which breaks the test for Python 3.11.4 released on June 6.

zhiruiluo: They will likely backport this change to all Python 3.11 verions.

test/test_decorator.py
```diff
@@ -43,6 +41,14 @@ def pop(self, index, default_value=None):
...
 
 
+def delay_evaluation_wrapper(fn: Callable) -> Callable:
+    def delayed_call(*args, **kwargs):
+        return lambda: fn(*args, **kwargs)
+
+    return delayed_call
+
+
+@delay_evaluation_wrapper
```
lebrice, line 51:
```suggestion
@functools.partial
```

lebrice (changes requested):
I don't really understand the change in the inspect module, and I also don't quite understand your changes here either.. 

I do understand that you want to delay the call to `change_signature(fn_with_all_argument_types, 1, c=1)` because executing it in Python 3.11.4 raises a `ValueError("non-default argument follows default argument")`.

However, before your change, `partial` here is meant as `change_signature`, it makes a variant of a function with a different signature (different default values in the signature).

Here your `delayed_wrapper` is exactly the same as `functools.partial`, and when applied to the `partial` (a.k.a. `change_signature`, it doesn't change anything, as far as I can tell. Why are you applying it to the `change_signature` function?

It also seems to my eyes like you are changing the behaviour of the rest of the test cases by only adding this additional level of "delay" to only _some_ of the test cases.


I'm going to re-read your PR, but I'd really appreciate if you could clarify this a bit for me.

test/test_decorator.py
lebrice: And at that point, as far as I can tell, this doesn't do anything, so why are we adding it?

lebrice:
Closed in favour of #273

PR #290, opened 2023-10-23
Disallow extra nonexistent fields

anivegesana:
If a config file is specified that contains extra keys that don't actually exist in the dataclass, those extra keys are ignored. I think it makes sense to give an error if additional fields are present in the config file when they don't actually exist.


simple_parsing/wrappers/dataclass_wrapper.py
```diff
@@ -295,17 +295,22 @@ def set_default(self, value: DataclassT | dict | None):
...
             field_wrapper.set_default(field_default_value)
+            unknown_names.remove(field_wrapper.name)
         for nested_dataclass_wrapper in self._children:
             if nested_dataclass_wrapper.name not in field_default_values:
                 continue
             field_default_value = field_default_values[nested_dataclass_wrapper.name]
             nested_dataclass_wrapper.set_default(field_default_value)
+            unknown_names.remove(nested_dataclass_wrapper.name)
+        unknown_names.discard("_type_")
+        assert not unknown_names, f"Fields {unknown_names} not found in {self.dest}"
```
lebrice, line 313: I'd preffer if we raised an error in this case:
```suggestion
        if unknown_names:
            raise RuntimeError(f"{sorted(unknown_names)} are not fields of {self.dataclass} at path {self.dest!r}!")
```

test/test_set_defaults.py
```diff
@@ -40,6 +40,21 @@ def test_set_defaults_from_file(tmp_path: Path):
...
+    parser.add_arguments(Foo, dest="foo")
+
+    saved_config = Foo(a=456, b="HOLA")
+    config_path = tmp_path / "broken_foo.yaml"
+    broken_yaml = to_dict(saved_config)
+    broken_yaml["i_do_not_exist"] = 3
+    with open(config_path, "w") as f:
+        yaml.dump({"foo": broken_yaml}, f)
+
+    with pytest.raises(AssertionError):
```
lebrice, line 54:
```suggestion
    with pytest.raises(
        RuntimeError,
        match=(
            r"\['i_do_not_exist'\] are not fields of <class 'test.test_set_defaults.Foo'> at path 'foo'!"
        ),
    ):
```

lebrice (approved):
Thanks @anivegesana , Cool idea, I like this "structured config" style of improvement.

I left some minor comments.

Issue #297, opened 2023-12-28
Union of singular type with list of types not behaving as expected.


rwightman:
**Describe the bug**
So this may not be a bug, but it's not clear what the expected behaviour is in any case. For class fields that are often singular but sometimes a list or tuple of values of the same type I often use `Union[type, List[type]]` in annotations. When it's a single value I access `obj.value` instead of `obj.value[0]`, there is code to check for sequences and handle appropriately.

I thought this might just work with SimpleParsing, it accepts the type definition but does not behave the way I was hoping, and I cannot determine what it is supposed to do in this case, it looks like it ignores the list/tuple typing and behaves as if the field was just `type`

**To Reproduce**
```python
from simple_parsing import ArgumentParser
from dataclasses import dataclass

@dataclass
class Foo:
   bar: Union[str, List[str]]

if __name__ == "__main__":
   parser = ArgumentParser()
   parser.add_arguments(Foo, "foo")
   args = parser.parse_args()
   foo: Foo = args.foo
   print(foo)
```

**Expected behavior**

```console
$ python issue.py --bar alpha
Foo(bar='alpha')

$ python issue.py --bar alpha beta
Foo(bar=['alpha', 'beta'])
```

**Actual behavior**
A clear and concise description of what is happening.

```console
$ python issue.py --bar alpha
Foo(bar='alpha')

$ python issue.py --bar alpha beta
error: unrecognized arguments: beta
```

**Desktop (please complete the following information):**
 - Version: 0.1.4
 - Python version: 3.11

**Additional context**

lebrice:
Hey there @rwightman , thanks for posting!


Hmmm this is clearly a bug. I'm curious though, does this happen only with strings? Or do other field types exhibit the same behaviour?

rwightman:
@lebrice it behaves similarly if I change str -> int in this example

grahamannett:
yeah i just ran into this as well.  

trying to understand exactly why/where it happens but not sure how much more time i will spend on this.  best guess right now is that it happens from the `wrapped_dataclass.add_arguments(parser=self)` call in parsing.py if anyone else runs into this issue and wants a starting point.

grahamannett:
found this as well and dove into it a bit more (from an earlier comment I deleted).  

[...]

PR #308, opened 2024-03-13
Union handles lists

grahamannett:
For #297

There might be a more elegant way to handle this but I think it would require redoing a lot more than this?

I also am not sure if this is the correct place to put this as there seemed like 2-3 other potential places and maybe more if i kept looking.

The if/else stuff is basically just trying to handle the equivalent of below (which probably makes more sense if there are more cases  to worry about for UnionTypes):

```python

  if any(utils.is_list(o) for o in utils.get_args(self.type)):
      # for: Union[list[int], list[str]] keep as list
      if all(utils.is_list(o) for o in utils.get_args(self.type)):
          return raw_parsed_value
      # for Union[str, list[str]] if single value, then dont keep as list
      elif len(raw_parsed_value) == 1:
          return raw_parsed_value[0]
  # are there other cases?
```

### DIFF
--- a/simple_parsing/wrappers/field_wrapper.py
+++ b/simple_parsing/wrappers/field_wrapper.py
@@ -330,6 +330,8 @@ class FieldWrapper(Wrapper):
         elif self.is_union:
             logger.debug("Parsing a Union type!")
             _arg_options["type"] = get_parsing_fn(self.type)
+            if any(utils.is_list(o) for o in utils.get_args(self.type)):
+                _arg_options["nargs"] = "*"
 
         elif self.is_enum:
             logger.debug(f"Adding an Enum attribute '{self.name}'")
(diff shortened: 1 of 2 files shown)


lebrice (approved):
Nice! Thanks @grahamannett ! :)

grahamannett:
Seems like the linter or pre-commit checks failed? Not sure if that was because of me, I didn't see any info on what to do and the default hooks/pre-commit stuff did not work for me but didn't spend too much time trying.  

I can go back and try and fix the PR to see if that would be helpful and besides line length I am not sure what other things in the pre-commit might be different (I use black but probably should switch to ruff it seems, just had issues getting it to work as I wanted when I had looked into using it prior)

lebrice:
Hey @grahamannett , yes its the pre-commit hooks which are failing. 
Can you do 
```
pip install pre-commit
pre-commit run --all-files
``` 
And commit the results? That should fix it.

Thanks again!

grahamannett:
@lebrice just ran it and commit the file that changed related to what I PR'ed earlier but still says ruff-format is failing from my setup.

lebrice:
> @lebrice just ran it and commit the file that changed related to what I PR'ed earlier but still says ruff-format is failing from my setup.  

Ok no prob I'll take a look tomorrow.
Thanks again!

PR #318, opened 2024-06-28
Try to use stdlib tomllib before tomli

bkad:
Since Python 3.11, the standard library comes with a [TOML parser](https://docs.python.org/3/library/tomllib.html#module-tomllib).


bkad:
That's true, although tomli already specified as optional so I don't think it would functionally change much.

lebrice:
I'd like to merge this, but I think there's something weird going on with the CI, it didnt run any checks here.. I'll investigate.

lebrice (approved):

lebrice:
Testing something here, will close and re-open to try to trigger the CI workflow.

PR #323, opened 2024-08-27
Fix is_optional(Literal[None])

krzentner:
Without this, a field with a type like Literal[None, "a", "b"] may be considered required on some code paths. In particular, when specifying a default for an argument group. Note that this bug is difficult to work around because None | Literal["a", "b"] and Optional[Literal["a", "b"]] result in different (worse) parsing behavior.


lebrice:
Awesome! Thanks a lot @krzentner !

lebrice (approved):

PR #334, opened 2024-12-02
Add option to specify config path argument name

joshlk:
Add an option so the user can specify the name of the config path argument.

I have extended the `add_config_path_arg` parameter of `ArgumentParser` to accept a string. If specified this determines the name of the argument, otherwise, it defaults to the previous value of `config_path`.

I have also added the `config_path` parameter to `ArgumentParser`'s docstrings, as it was missing previously.

### DIFF
--- a/simple_parsing/parsing.py
+++ b/simple_parsing/parsing.py
@@ -166,11 +166,6 @@ class ArgumentParser(argparse.ArgumentParser):
         if add_config_path_arg is None:
             # By default, add a config path argument if a config path was passed.
             add_config_path_arg = bool(config_path)
-        if isinstance(add_config_path_arg, str) and not add_config_path_arg.isidentifier():
-            raise ValueError(
-                "If `add_config_path_arg` is a string it must be a valid Python identifier (no dashes)."
-                f" Not: {add_config_path_arg}"
-            )
         self.add_config_path_arg = add_config_path_arg
 
     # TODO: Remove, since the base class already has nicer type hints.
@@ -330,7 +325,7 @@ class ArgumentParser(argparse.ArgumentParser):
                 help="Path to a config file containing default values to use.",
             )
             args_with_config_path, args = temp_parser.parse_known_args(args)
-            config_path = getattr(args_with_config_path, config_path_name)
+            config_path = getattr(args_with_config_path, config_path_name.replace("-", "_"))
 
             if config_path is not None:
                 config_paths = config_path if isinstance(config_path, list) else [config_path]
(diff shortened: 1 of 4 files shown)


simple_parsing/parsing.py
```diff
@@ -159,6 +166,11 @@ def __init__(
         if add_config_path_arg is None:
             # By default, add a config path argument if a config path was passed.
             add_config_path_arg = bool(config_path)
+        if isinstance(add_config_path_arg, str) and not add_config_path_arg.isidentifier():
+            raise ValueError(
+                "If `add_config_path_arg` is a string it must be a valid Python identifier (no dashes)."
+                f" Not: {add_config_path_arg}"
+            )
```
lebrice, line 173: Actually, you can use attribute names with dashes when you `getattr` and `setattr` on a `Namespace` object:
```python
>>> from argparse import Namespace
>>> ns = Namespace()
>>> setattr(ns, "some-attr", 42)
>>> getattr(ns, "some-attr")
42
```

simple_parsing/parsing.py
```diff
@@ -1010,10 +1027,12 @@ def parse(
 
     If `config_path` is passed, loads the values from that file and uses them as defaults.
     """
+    if dest == add_config_path_arg:
+        raise ValueError("`add_config_path_arg` cannot use the same name as `dest`.")
```
lebrice, line 1031:
```suggestion
        raise ValueError("`add_config_path_arg` cannot be the same name as `dest`.")
```

lebrice (approved):
Thanks a lot @joshlk , this looks good! There's just a minor comment here about the extra restriction that you place on the values that `config_path_name` can take.

simple_parsing/parsing.py
joshlk: > Actually, you can use attribute names with dashes when you `getattr` and `setattr` on a `Namespace` object:
> 
> ```python
> >>> from argparse import Namespace
> >>> ns = Namespace()
> >>> setattr(ns, "some-attr", 42)
> >>> getattr(ns, "some-attr")
> 42
> ```

Thanks good idea. I have it working now. For reference, you have to replace the dashes with underscores like so:
```python
getattr(args_with_config_path, config_path_name.replace('-', '_'))
```

joshlk: > Also, if you have the time, would you mind adding a simple test so we can cover this change?
> 
> ```python
> @pytest.mark.parametrize("invalid_value", ["foo.bar.baz?", "-------", "bob bob bob"])  # add other more relevant examples perhaps?
> def test_pass_invalid_value_to_add_config_path_arg(invalid_value):
>     with pytest.raises(ValueError):
>         _ = ArgumentParser(add_config_path_arg=invalid_value)
> ```

I've added some tests

joshlk:
@lebrice I've addressed the comments. Its ready for a re-review

joshlk:
Hi @lebrice, is this PR ready to be merged?

lebrice:
Hi @joshlk let me take a quick look. I'll get back to you.

lebrice:
Ok all good, thanks @joshlk !

Issue #331, opened 2024-10-03
Bug with subgroups and config file


farzadab:
**Describe the bug**
Subgroups work if parameters are provided as CLI args, but it fails if the same parameter is passed through a config file. See below:

**To Reproduce**
```python
from typing import Union
import dataclasses
import simple_parsing

@dataclasses.dataclass
class ModelTypeA:
    model_a_param: str

@dataclasses.dataclass
class ModelTypeB:
    model_b_param: str

@dataclasses.dataclass
class TrainConfig:
    model_type: Union[ModelTypeA, ModelTypeB] = simple_parsing.subgroups(
        {"type_a": ModelTypeA, "type_b": ModelTypeB},
        default_factory=ModelTypeA,
        positional=False,
    )


# this works:
simple_parsing.parse(config_class=TrainConfig, args=['--model_a_param', 'test'])

# this doesn't work and throws an error
simple_parsing.parse(config_class=TrainConfig, add_config_path_arg=True, args=['--config_path', 'config.yaml'])

# config.yaml:
# model_a_param: test
```

**Expected behavior**
I expect the config file to be allow the same behaviour as the CLI args.

**Actual behavior**
As described above, the second invocation leads to the following error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/farzad/Library/Caches/pypoetry/virtualenvs/ultravox-o1nHeuex-py3.11/lib/python3.11/site-packages/simple_parsing/parsing.py", line 1030, in parse
    parsed_args = parser.parse_args(args)
                  ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/farzad/.pyenv/versions/3.11.4/lib/python3.11/argparse.py", line 1869, in parse_args
    args, argv = self.parse_known_args(args, namespace)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/farzad/Library/Caches/pypoetry/virtualenvs/ultravox-o1nHeuex-py3.11/lib/python3.11/site-packages/simple_parsing/parsing.py", line 321, in parse_known_args
    self.set_defaults(config_file)
  File "/Users/farzad/Library/Caches/pypoetry/virtualenvs/ultravox-o1nHeuex-py3.11/lib/python3.11/site-packages/simple_parsing/parsing.py", line 405, in set_defaults
    wrapper.set_default(default_for_dataclass)
[...]

PR #343, opened 2025-02-26
Fix subgroup

tcapelle:
This is a standing issue fix: #331 

  We successfully fixed the issue where subgroups didn't work correctly when parameters were provided through config files. The core problem was that when subgroup parameters were specified in a YAML file, they
  couldn't be properly associated with their parent subgroup type.

  The Fix

  The solution involved two main components:

  1. Smart field association for config files:
    - Added code to detect when fields in a config file match fields in a subgroup's default type
    - When matching fields are found, they're automatically associated with the correct subgroup
    - This happens in both dataclass_wrapper.py and parsing.py when processing configs
  2. Improved callable type resolution:
    - Enhanced how we handle functions and functools.partial objects in subgroups
    - Added robust resolution of return type annotations to better support nested class definitions
    - Fixed edge cases with string annotations by evaluating them in the correct scope

### Before:
```
# This would fail with "RuntimeError: ['model_a_param'] are not fields of <class 'TrainConfig'>"
```
```yaml
model_a_param: test
```
### After:
Both approaches now work equivalently:

CLI args
```
  simple_parsing.parse(TrainConfig, args=['--model_a_param', 'test'])
```
Config file
```  
simple_parsing.parse(TrainConfig, config_path='config.yaml')  # where config.yaml contains 'model_a_param: test'
```
The fix preserves backward compatibility while adding the ability to specify subgroup parameters directly in config files without explicit nesting, making the API more intuitive and consistent.

Mostly authored by Claude Code 3.7

### DIFF
--- a/simple_parsing/parsing.py
+++ b/simple_parsing/parsing.py
@@ -1172,7 +1172,47 @@ def _create_dataclass_instance(
                 choices = field_wrapper.subgroup_choices
                 default_factory = choices[default_key]
                 if callable(default_factory):
-                    default_type = default_factory
+                    # Handle callables (functions or partial) that return a dataclass
+                    if isinstance(default_factory, functools.partial):
+                        # For partial, get the underlying function/class
+                        default_type = default_factory.func
+                    else:
+                        default_type = default_factory
+                    
+                    # If it's still a callable but not a class, we need the return type
+                    if not isinstance(default_type, type) and callable(default_type):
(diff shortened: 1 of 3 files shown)


lebrice:
Hey there @tcapelle , thanks a lot for this!

I'm going to take a look at the code and I'll get back to you. Also, do you mind if I push some changes to your branch?

tcapelle:
go ahead! thanks for looking at this.

PR #345, opened 2025-03-04
Bugfix: fix optional positional arguments.

joshlk:
Optional positional argument previously raised an error with argparse:

```
TypeError: 'required' is an invalid argument for positionals
```


lebrice (approved):
Awesome, neat pr @joshlk , thanks a lot!

PR #354, opened 2025-05-29
Feature: able to parse `Optional[Type]` field to `None` from command-line

anqixu:
Currently, if a field with type-hint `Optional[Type]` has a non-optional default value, there's no way to set it to `None` by parsing from the command line. E.g.

```
@dataclass
class Config2:
    var_with_default: Optional[int] = 42
```

```
$ ./run.py --var_with_default None
```

This PR adds this feature, specifically by parsing the string literal `"None"` into` None` for optional-typed fields.


anqixu:
@lebrice could you kindly have a review of the updated PR? Previous version failed CI due to silly issue (comparing type() directly), which I since fixed.

PR #362, opened 2026-01-26
fix: explicitely set version to 0.1.8 in pyproject.toml

naetherm:
Add required version setting, otherwise can lead to build problems:
```
The Poetry configuration is invalid:
  - Either [project.version] or [tool.poetry.version] is required in package mode.
```


lebrice:
Hello! This now uses [uv](docs.astral.sh/uv) instead of Poetry, and the version is dynamically determined based on the tags and commits, so it shouldn't be hard-coded in the pyproject.toml file.

Thanks anyway @naetherm ! :)

naetherm:
Oh, indeed! Then it gets time to update that [on my side](https://aur.archlinux.org/packages/python-simple_parsing) :)

